PR #36, opened 2015-02-19
Include package data

pumazi:


karenc:
How about just not including the tests in the source distribution?

PR #50, opened 2016-02-24
Use cnx-archive-uri in epub as the ident_hash of the adapted items

karenc:


cnxepub/adapters.py
```diff
@@ -233,16 +233,25 @@ def _node_to_model(tree_or_item, package, parent=None,
     return result
 
 
-def _id_from_metadata(metadata):
-    """Given an item's metadata, discover the id."""
+def _ident_hash_from_metadata(metadata):
```
pumazi: From the perspective on cnx-epub this is an id, not an ident_hash. If another system is to use this library, they could use things like `abc123` as the identifier. We don't want to enforce our internal identifier on anyone at this layer.

karenc: You're right.

PR #51, opened 2016-03-01
Add DocumentContentFormatter and HTMLFormatter

karenc:


cnxepub/adapters.py
```diff
@@ -313,6 +310,20 @@ def __init__(self, item, package):
         self.resources = resources
 
 
+class DocumentContentFormatter(object):
+    def __init__(self, document):
+        self.document = document
+
+    def __str__(self):
```
karenc: Oops, there might be a problem using `__str__`.  Looking into it.

pumazi:
Nice work! :100: This is well on the way to making things better. :+1:

PR #54, opened 2016-03-03
Add script for generating a single HTML file from an epub

karenc:
Example usage:

```
cnx-epub-single_html e79ffde3-7fb4-4af3-9ec8-df648b391597@7.1.epub physics.html
```

A rewrite of
https://github.com/Connexions/cte/blob/8b894d723d4e184f998feedf7530898055935329/cte-collation-poc/fullbook.py


reedstrm:
This looks great, except for the Q. of code location. Clearly the formatter goes here, but does the CLI script itself? It depends on the existence of an archive, so it seems to me that it should live in either cnx-archive (fits the read-only mandate there) or cnx-publish, since this is a commandline tool to use for getting a file to do other publish things with. Not a huge thing, since it's mostly just to provide a visible wrapper around the formatter. I'm tempted to say we need a cnx-utils repo, with _just_ clis that for example our textbookdev people could install. They have no need for a local copy of any of archive publish or epub, they just want the 'grab an full HTML doc, just like what transforms will apply to' functionality.

pumazi:
@reedstrm in response to you location question...

I believe this is the correct location. I'm actually putting the cooking integration bits here too. The reason for that is because, all the data is encapsulated and only depends on the EPUB structure, not archive or publishing. (I personally think we've got this spot on! Huge win! :100: )

To clarify, this code does require the models, and thus is dependent on cnx-epub. So, your proposal to split the package may be good, but probably fragments too much. In the future, installing cnx-epub alone should provide the textbook devs with the tools they require to work with the content. For example, `pip install cnx-epub[collation]` should install all the cnx-epub and the necessary dependencies to build a collated book and write it back out to internal epub. And all that can be done without interacting with archive (or whatever repo system has the canonical data).

reedstrm:
o.k., so it extends cnx-epub into "models and tools for epubs" Yup, that works. I'll merge.

PR #56, opened 2016-03-04
List resources in metadata in epub html only if there are resources

karenc:


cnxepub/tests/scripts/test_single_html.py
```diff
@@ -24,7 +24,7 @@ def target(self):
 
     epub_path = os.path.join(TEST_DATA_DIR, 'book')
 
-    maxDiff = 7000
+    maxDiff = 30000
```
pumazi: For future reference, `maxDiff = None` is best.

PR #58, opened 2016-03-08
Html parsers changes

karenc:
- Change html metadata parser to use relative xpath
    
    This allows parsing a html file with more than one metadata element.
    For example, in the single html format for binders, there is one
    metadata element per document in the book.

  - Change html metadata description parser to not return the div
    
    Instead of returning:
    
    ```html
    <div class="description" itemprop="description" data-type="description">
      The summary:
      <p>summary</p>
    </div>
    ```
    
    return:
    ```html
    The summary:
    <p>summary</p>
    ```


pumazi:
Have you tested this in conjunction with cnx-publishing?

karenc:
No, I'll do that.

karenc:
I fixed some bugs while running the publishing tests:

```diff
diff --git a/cnxepub/html_parsers.py b/cnxepub/html_parsers.py
index 772a8fc..df64e98 100644
--- a/cnxepub/html_parsers.py
+++ b/cnxepub/html_parsers.py
@@ -134,10 +134,8 @@ class DocumentMetadataParser:
             description = items[0]
             value = [description.text]
             for child in description.getchildren():
-                value.append(child.text)
                 value.append(etree.tostring(child).decode('utf-8'))
                 value.append(child.tail)
-            value.append(description.tail)
             value = ''.join(value).encode('utf-8')
         except IndexError:
             value = None
```

karenc:
I'm not sure if this makes sense :confused:  I've added a `DocumentSummaryFormatter`.

pumazi:
What's the rational for the description change?

The relative xpath change makes perfect sense.

karenc:
We keep returning an additional div wrapper when we parse the summary / description in the html.

pumazi:
I suppose this is necessary, but I don't like that it's a formatter, because we don't have a model that represents the summary. The formatters as they are at this time are simply wrapping classes, but if you think of them in terms of interface adaptation, there is no interface for the summary. Currently, this concept of the formatter can be refactored to something like `IHTMLFormatter(doc)` where 'doc' implements `IDocument` and `SingleHTMLFormatter` implements `IHTMLFormatter` and has been registered to adapt `IDocument`. We don't see that story for the summary/description data, but that doesn't mean it couldn't exist.

I suppose that's too much forward thinking and not enough thinking in the reality of the situation.

PR #61, opened 2016-04-01
Collate

karenc:


karenc:
\o/ It's working!  Please do `git rebase --autosquash -i master` before merging.

cnxepub/formatters.py
```diff
@@ -191,7 +191,7 @@ def __str__(self):
     def __bytes__(self):
         if not self.built:
             self.build()
-        return etree.tostring(self.root, pretty_print=True)
+        return etree.tostring(self.root, pretty_print=True, method='html')
```
karenc: This should fix the weird nesting problem because the browser doesn't see the file as xhtml if the file is saved with extension `.html`

screen shot 2016-03-31 at 16 13 25

PR #64, opened 2016-04-18
Integrates cnx-easybake

pumazi:


cnxepub/tests/test_collation.py
```diff
@@ -239,3 +239,45 @@ def test_without_ruleset(self):
...
+                              'title': "Document One"}),
+                self.make_document(
+                    id="3c448dc6",
+                    content=b"<span>document two</span>",
+                    metadata={'version': '1',
+                              'title': "Document Two"})])
+
+        # Append a ruleset to the binder.
+        ruleset = io.BytesIO(b"""\
+div[data-type='metadata']:last-of-type {
```
reedstrm: Hmm, does this work? I have a  feeling :last-of-type may not be implemented in cssselect2, yet.

pumazi: Yeah, you're right. You can see this in the test, since 'Document One' is the value for title.

pumazi: BTW, the `:last-of-type` does seem to work. However, I can't seem to get it to work correctly in the context of xhtml.

cnxepub/tests/test_collation.py
```diff
@@ -280,4 +280,4 @@ def test_with_ruleset(self):
         # Check for the appended composite document
         self.assertEqual(len(collated_binder), 3)
         self.assertEqual(collated_binder[2].metadata['title'],
-                         'Composite One')
+                         'Document One')
```
karenc: Hm... I'm not sure if I understand this commit.  Why are we not getting the titles from the TOC as the title overrides?

pumazi: As I wrote in the extended commit message, 
> Since the single-page is not a cannonical source and does not need to retain the complete information of the document, the title overrides in this case, ignoring the TOC during reconstitution is okay.

We don't have the ability to do this kind of creating and movement of TOC items in the ruleset just yet. Plus, during this collation process, the title of the document will always equal the TOC item title, because the title-override is used during rendering.

karenc: I read your commit message but I needed some extra explanation.  I understand that the composite pages will not have title overrides but what about normal pages?

pumazi: Because they are rendered out to html, and the html renderer uses the title-override rather than the document's title, the pages will contain the correct title in both places.

karenc: The problem appears to be, if we don't ignore the TOC, easybake is not going to work.  I recommend adding `FIXME` and commenting out the code instead of removing it.

pumazi:
Issue #65 was created to capture the title-overrides issue.

.travis.yml
```diff
@@ -7,6 +7,10 @@ before_install:
   # Install the coverage utility and coveralls reporting utility
   - pip install coverage
   - pip install coveralls
+  # FIXME (18-Apr-2016) cnx-easybake requires an unreleased package, cssselect2
+  - pip install git+https://github.com/SimonSapin/cssselect2.git#egg=cssselect2
```
reedstrm: I've forked cssselect2 into the Connexions space, in prep for taking it over, so this could switch to:
git+https://github.com/Connexions/cssselect2.git#egg=cssselect2

PR #69, opened 2016-05-11
Book zip

cburk:
Book zip tool.  Still needs review from @reedstrm


cburk:
Another question, there are some downloads that need to occur in order for this to run (for example zip).  Is there a standard way to do this?

pumazi:
Hello @cburk, I don't believe this is the correct place for this logic to live. You should consult with @reedstrm about where this should go. Thanks. :)

edwoodward:
We told him to put it here but that was a couple of months ago.
On May 11, 2016 5:01 PM, Michael Mulich <notifications@github.com> wrote:Hello @cburk, I don't believe this is the correct place for this logic to live. You should consult with @reedstrm about where this should go. Thanks. :)

—You are receiving this because you are subscribed to this thread.Reply to this email directly or view it on GitHub


  
  





!DSPAM:1106,5733abb9298781976015428!

edwoodward:
Let's discuss where this should live if this is not the right place before we close the PR.

karenc:
What is this script used for?  I don't think it fits into any of the backend packages.  Do we have a scripts repository?

edwoodward:
It is pulling HTML from archive and creating a zip file that can be used for viewing the content offline. The DMS team requested this.

karenc:
Oh I see, does the DMS team have a repository for this kind of script?

edwoodward:
Possibly. I'll test his script and try to determine where it should live.

philschatz:
This seems to have similar goals as the EPUB format (viewable offline). Would it make sense to have a script that builds the EPUB by pointing to an archive URL? Then it could be used to generate an EPUB (which works with the publishing) and be able to view it?

As an example, there's: https://github.com/oerpub/demo-book/

and the TOC file: https://github.com/oerpub/demo-book/blob/gh-pages/content/siyavula-physics-12-nav.html

And viewing the HTML (for DMS's) using `gh-pages` (no magic): 

- physics https://oerpub.github.io/demo-book/content/ieb-physics-skeleton-nav.html
- phyics2 https://oerpub.github.io/demo-book/content/siyavula-physics-12-nav.html

**Aside:** You _can_ include more than one book (like Economics, MicroEconomics, MacroEconomics) but don't have to.

edwoodward:
The need from the DMS team was to take the zip and upload it to a 
College LMS. They have been converting our non-working complete zip to 
do this and it is very labor intensive. This zip should make it easier.

Ed

pumazi:
Can someone point me to the use-case document for this work? I'd like to be a bit more informed about the problem this is attempted to resolve.

edwoodward:
This is the document I wrote for Christian - 
https://docs.google.com/document/d/1maXhsBjz0YUh21rCv7DjILA6G0jxwyT2C8An5b6OTbc/edit#heading=h.c59yb4mfjooy

philschatz:
In that case, would just `wget -m http://archive.cnx.org/contents/031da8d3-b525-429c-80cf-6c8ed997733a@9.30` work (download the HTML + images of the book)? There's an additional command-line arg to rewrite the links to be relative `../[UUID].html` instead of `/contents/[UUID].html`

Tested just now and it downloads all the HTML files as well as images.

pumazi:
IMHO the EPUB format does much of what you've outlined in your document. I'll admit, we could make it more human friendly, but this hasn't been an area we've had time to work on just yet.

I don't think it's a very good idea to create yet another format that we need to maintain. Is there anyway we can restart this discussion in terms of how can we use our existing (and accepted) format to fulfill this use-case?

Also, this work seems similar to something we have in the CTE backlog. So there may be room for making this more official. ( CC @brucep99166 )

BruceP99166:
TEA requirements will be a factor.  @nyxer95 as Product Manager is the one to identify requirements and priority for this work.

PR #70, opened 2016-05-17
Add auto generated ids to single html

karenc:


cnxepub/models.py
```diff
@@ -467,6 +486,58 @@ def references(self):
             return []
         return self._references
 
+    def fix_html_ids(self, mapping):
```
karenc: These probably don't belong to the models ... :/
May be I need to put everything into the single html formatter?

cnxepub/tests/test_models.py
```diff
@@ -357,3 +357,52 @@ def test_document_content(self):
...
+<p class="para"><a href="/contents/{other_page}#title">Link</a> to other page title</p>\
+""".format(other_page=page_one_id)
+
+        new_frag_id = 'id-{}-1'.format(page_one_id)
+        expected_content2 = """\
+<div class="title" id="id-{id}-1">Page two</div>
+
+<p class="para" id="id-{id}-2">This thing and <em>that</em> thing.</p>
+
+<p class="para" id="id-{id}-3"><a href="/contents/{other_page}#{new_frag_id}">Link</a> to other page title</p>\
```
pumazi: Aha, here appears to be the important bit. It looks like it works.

cnxepub/models.py
pumazi: Do you feel you can do that?

I suppose if we view ids as equal members of the data to that of resource links, then having it here on the model is probably alright.

reedstrm: The presence of IDs on specific tags is a feature of of HTML5 dialect, so arguably is part of our model specification. AS you point out, it's hard to link without them. :-)

cnxepub/models.py
```diff
@@ -467,6 +486,58 @@ def references(self):
...
+        xpath = '|'.join([elements_xpath, data_types_xpath])
+
+        mapping = {}  # old id -> new id
+
+        for i, node in enumerate(self._xml.xpath(xpath)):
+            old_id = node.attrib.get('id')
+            new_id = 'id-{}-{}'.format(self.id, i)
+            node.attrib['id'] = new_id
+            if old_id:
+                mapping[old_id] = new_id
```
reedstrm: Ah, so we're guaranteeing uniqueness by replacing all the existing ids. Hmm, the existing IDs are potentially exposed as part of the user content, however - there could be other pages linking to specific content on this page i.e. 'See <a href="/contents/b619e160-1c44-11e6-afac-0090f5e17687@4#table-3">Table of Symbols</a>' I realize this is spec/scope creep, but I do think we have to leave existing IDs in place (though I suppose we can replace any that are _already_ duplicates ... I'm sure there are plenty!) So, Id' like to see a step that grabs all the existing ids in the doc, then iterates through linkable types and only adds an id to those that do not have one already, and tests each new id for existence before assigning it. Dups could be handled while building the first list, by removing their ID (first one gets to keep it). This allows us to not have to rewrite internal links at all, since we can't rewrite all the potential external links that may exist.

reedstrm: Just had a discussion w/ Ed, and I _think_ I came to a conclusion that we _can_ replace the IDs, _in the cooked content_. We currently have lots of ID based problems with editing (deriving copies, combining pre-existing pages, etc.)

karenc: At the moment, this (id generation) code is only invoked when running the single html script.

karenc:
Going to change single html formatter and adapter to generate unique IDs by prepending the page ID in the formatter and when we adapt the html back into models, we'll change the IDs back because IDs only need to be unique within the page.

reedstrm:
Excellent - and inside easybake, when I copy nodes, I'll add a postfix, so the new nodes will have the expected prefix, and will get stripped when adapting back, to 'copied-id+postfix'

karenc:
I've updated the branch but github.com doesn't seem to be showing it yet...

reedstrm:
rebased to master

karenc:
This is ready for review again.

PR #71, opened 2016-05-18
Get the title overrides from TOC when adapting single html

karenc:
Depends on https://github.com/Connexions/cnx-easybake/pull/15

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -335,9 +335,8 @@ def adapt_single_html(html):
     body = html_root.xpath('//xhtml:body', namespaces=HTML_DOCUMENT_NAMESPACES)
     _adapt_single_html_tree(binder, body[0])
 
-    # FIXME (20-Apr-2016) https://github.com/Connexions/cnx-epub/issues/65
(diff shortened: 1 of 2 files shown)


reedstrm:
I note that this used to run off the end of the list. Does it depend on the struture of the incoming HTML at all> i.e. is is possible to cause that same error by writing a bad ruleset-transform?

karenc:
If the table of contents is not updated, we will have that index error again...  Perhaps I should change this to be more robust.

reedstrm:
Yeah, if it falls off the end, perhaps set it to something slightly generic '(untitled)' or something?
(we block that specific string in legacy from being published, BTW)

karenc:
If the table of contents is not the same length as the number of pages in the book, should it just not use title overrides?

karenc:
There's a possibility that we have a composite page in the middle of the book and if we just add "(untitled)" to the pages at the end of the book, the title overrides might be shifted by one?

reedstrm:
Yes, that's true - it will happen, we're adding pages at the end of each chapter. You're right, if we error, we should back out of the whole thing.

PR #72, opened 2016-05-19
Don't use title overrides if some are missing

karenc:
Need tests

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -47,6 +47,10 @@ __all__ = (
     )
 
 
+class AdaptationError(Exception):
+    """Raised when data is not able to be adapted to the requested format."""
+
+
 def adapt_package(package):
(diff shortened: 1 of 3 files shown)


karenc:
Need more tests.

reedstrm:
I think this is probably wrong. We _should_ break in this case. Any objections to closing this?

karenc:
Does it mean we don't need to handle this case? Just close this and delete the branch?

reedstrm:
Hmm, looking at the actual code, I think perhaps detect the mismatch and error appropriately, rather than throwing a mysterious IndexError from the loop.

karenc:
@reedstrm do you plan to add tests?

reedstrm:
Test added, and it seems to be passing. Yay!

karenc:
Test looks good, thanks!  Do you mind squashing the first 3 commits?

PR #73, opened 2016-05-19
Html in titles

pumazi:
This is logic to fix @helenemccarron's issue on cte-dev. The titles within her nav tree contain an outer wrapping div as well as inner html. At this time we are going to allow the outer wrapping div tag, but it will be removed in the future.

### DIFF
--- a/cnxepub/html_parsers.py
+++ b/cnxepub/html_parsers.py
@@ -16,6 +16,15 @@ HTML_DOCUMENT_NAMESPACES = {
     }
 
 
+def _squash_to_text(elm):
+    value = [elm.text or '']
+    for child in elm.getchildren():
+        value.append(etree.tostring(child).decode('utf-8'))
+        value.append(child.tail or '')
+    value = ''.join(value).encode('utf-8')
+    return value
+
+
 def parse_navigation_html_to_tree(html, id):
     """Parse the given ``html`` (an etree object) to a tree.
     The ``id`` is required in order to assign the top-level tree id value.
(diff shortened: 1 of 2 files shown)


cnxepub/html_parsers.py
```diff
@@ -16,6 +16,15 @@
...
 
+def _squash_to_text(elm):
+    value = [elm.text or '']
+    for child in elm.getchildren():
+        value.append(etree.tostring(child).decode('utf-8'))
+        value.append(child.tail or '')
+    value = ''.join(value).encode('utf-8')
+    return value
+
+
```
reedstrm: this is basically the same as etree.tostring(elm, method=text, with_tail=False) no?

reedstrm: hos does this differ from etree.tostring(elm) ?

karenc: That was our first attempt.  There seems to be some problem with dealing with unicode / non-ascii characters.

reedstrm: >>> e.text=u'This is élément'
>>> etree.tostring(e1)
'<div><a>This is &#233;l&#233;ment</a></div>'
>>> etree.tostring(e1, method='html', encoding='utf-8')
'<div><a>This is \xc3\xa9l\xc3\xa9ment</a></div>'

One of those should work, no?

reedstrm: Existing content is only one deep, so I guess this'll work.

karenc: One of the tests failed last time we tried. Maybe there was a different
problem... Do you want to try running the tests with what you think will
work?

PR #74, opened 2016-05-25
Add a script for validating collated single-html files

pumazi:


pumazi:
I'll look into the failing tests... 😸

cnxepub/scripts/collated_single_html/main.py
```diff
@@ -5,31 +5,51 @@
...
+
     args = parser.parse_args(argv)
 
-    binder = cnxepub.adapt_single_html(args.collated_html.read())
-    pprint(cnxepub.model_to_tree(binder))
+    html = args.collated_html.read()
+    try:
+        htree = etree.fromstring(html)
+    except etree.XMLSyntaxError:
+        htree = etree.HTML(html)
```
karenc: Should this be inside `cnxepub.adapt_single_html`?

karenc: And should we use `cnxepub.reconstitute` instead of `cnxepub.adapt_single_html`? :D

pumazi: Is that a suggestion or question? I'd think you would know best, seeing as you wrote the `cnxepub.adapt_single_html` function.

pumazi: Ah, now that's an excellent idea. Perhaps adapt_single_html should remain xhtml specific at this time. Yeah, let's use `cnxepub.collation.reconstitute` and push this try..except there.

I'll take care of that change.

karenc: IMO there's no point in doing this in the validator script and then failing when we actually reconstitute.

pumazi:
⭐ Good work @karenc. Feel free to rebase it.

@reedstrm does this still work for you? If so, @karenc can move forward with the merge.

reedstrm:
I'll get Helene to test it. They've been using branch-as-of-10am successfully all day. :-)

cnxepub/scripts/collated_single_html/main.py
```diff
@@ -0,0 +1,52 @@
...
+"""Processes a collated book's single-html representation for validation.
+
+Ensures safe parsing for specific elements.
+"""
+from __future__ import print_function
+import argparse
+import logging
+import sys
+from pprint import pformat
+from lxml import etree
```
reedstrm: unused tree import, I think?

reedstrm:
Works on my test case cooked html.  Only niggle is the now unused etree import.

reedstrm:
I'm willing to merge, since all my code's been refactored away ... :-)

PR #75, opened 2016-05-26
make styles in single-html optional

reedstrm:
Helen's been editing these out, anyway.

### DIFF
--- a/cnxepub/scripts/single_html/main.py
+++ b/cnxepub/scripts/single_html/main.py
@@ -27,7 +27,7 @@ logger = logging.getLogger('single_html')
 
 
 def single_html(epub_file_path, html_out=sys.stdout, mathjax_version=None,
-                numchapters=None):
+                numchapters=None, styles=None):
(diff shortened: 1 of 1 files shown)


pumazi:
Removing the style altogether would probably be good. It's not really in the scope of the package to worry about style, no?

karenc:
I included the styles because it was in the original script that generates the single html in the cte repository.  I think it'd be better to remove them if Helene doesn't use them.

reedstrm:
I did have a thought of making it a variable that let's you point at a file of styles to incorporate, default none?

karenc:
Let's make it simpler and just remove the styles :D

reedstrm:
o.k. Could one of you do that, as you may have heard my laptop just died, and I don't have a handy copy of my ssh key ... (yikes!)

karenc:
Yes, I'll do that

karenc:
Remove styles in single html #76

Issue #78, opened 2016-06-17
Make sure html in titles format and adapt correctly from single HTML


karenc:
Part of https://github.com/Connexions/cnx-recipes/issues/34

For example, after collation, the title of a chapter may be 

```
<span>Chapter</span> <span>1</span> <span> | </span> <span>Sampling and Data</span>
```

Make sure this displays and is reconstituted correctly.

PR #79, opened 2016-06-17
Allow html in titles and correctly store all title overrides from single HTML

### LINKED ISSUE #77: Titles in nav are not being used for translucent binders when adapting from single HTML
All pages and chapters / units directly under the book have their title overrides stored, but everything in chapters / units have no title overrides.

Part of https://github.com/Connexions/cnx-recipes/issues/34

karenc:
Close #77 and #78.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -571,12 +571,14 @@ def html_listify(tree, root_xl_element, extensions, list_type='ol'):
     for node in tree:
         li_elm = etree.SubElement(root_xl_element, 'li')
         if node['id'] == 'subcol':
-            span_elm = etree.SubElement(li_elm, 'span')
-            span_elm.text = node['title']
+            span_elm = lxml.html.fragment_fromstring(
+                node['title'], create_parent='span')
+            li_elm.append(span_elm)
         else:
-            a_elm = etree.SubElement(li_elm, 'a')
-            a_elm.text = node['title']
+            a_elm = lxml.html.fragment_fromstring(
+                node['title'], create_parent='a')
             a_elm.set('href', ''.join([node['id'], extensions[node['id']]]))
+            li_elm.append(a_elm)
         if 'contents' in node:
             elm = etree.SubElement(li_elm, list_type)
             html_listify(node['contents'], elm, extensions)
--- a/cnxepub/html_parsers.py
+++ b/cnxepub/html_parsers.py
@@ -19,7 +19,7 @@ HTML_DOCUMENT_NAMESPACES = {
 def _squash_to_text(elm):
     value = [elm.text or '']
     for child in elm.getchildren():
-        value.append(etree.tostring(child).decode('utf-8'))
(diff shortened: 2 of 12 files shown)


cnxepub/formatters.py
```diff
@@ -571,12 +571,14 @@ def html_listify(tree, root_xl_element, extensions, list_type='ol'):
     for node in tree:
         li_elm = etree.SubElement(root_xl_element, 'li')
         if node['id'] == 'subcol':
-            span_elm = etree.SubElement(li_elm, 'span')
-            span_elm.text = node['title']
+            span_elm = lxml.html.fragment_fromstring(
+                node['title'], create_parent='span')
```
reedstrm: Won't this _force_ all titles to be HTML, and adds an extra wrapping <span> tag around all them, to boot?

cnxepub/html_parsers.py
```diff
@@ -16,11 +18,13 @@
...
 
-def _squash_to_text(elm):
+def _squash_to_text(elm, remove_namespaces=False):
     value = [elm.text or '']
     for child in elm.getchildren():
-        value.append(etree.tostring(child).decode('utf-8'))
+        value.append(etree.tostring(child).decode('utf-8').strip())
         value.append(child.tail or '')
+    if remove_namespaces:
+        value = [re.sub(' xmlns:?[^=]*="[^"]*"', '', v) for v in value]
```
reedstrm: sniff sniff - processing structured data w/ regexs usually means something is very wrong ... This also only allows for one level of markup in titles.

cnxepub/formatters.py
karenc: No, `span_elm = etree.SubElement(li_elm, 'span')`, we've always had a `<span>` around the title.

cnxepub/html_parsers.py
karenc: I don't see a way to remove namespaces.  If you don't like it I guess we don't have to remove the namespaces.

Why do you think this only allows one level of markup?

cnxepub/formatters.py
reedstrm: Ok. yes I see that now. Hmm. We may need to change that if Helene et al object, but it's the minimal change for now, agreed.

cnxepub/html_parsers.py
reedstrm: Ah, you're right it gets them all - I wasn't seeing recursion, but etree.tostring() does the entire subtree.

reedstrm: The problem with removing the namespace declarations like this is that we don't also remove the use of the namespace, we may end up with orphan tags. Admittedly, for our use-case the namespaces are almost always in the way - see my PR re: extending a bunch of xpaths to match both namespaced and non-namespaced (xhtml + html). The only trick I know for stripping them in a structured way is to create a root node with the default namespace set to what you want to remove, parse w/ ns_clean=True, then ... hehe I think it ends up using string manipulation in the end, to remove the default xmlns declaration. Hmm.

PR #80, opened 2016-06-21
detect and pass through attributes on root-of-document nodes

reedstrm:
Fairly invasive, hacky approach to passing through any attributes that are set on the root of the document node. Considering how often we're wrapping and unwrapping the Document tree, I think we may actually have a wrong model for Document.content: right now it's a HTML fragment,  I think we should just bite the bullet and refactor this, eventually. For now, I've added special case code to grab any non-template-added attributes (not itemscope/itemtype) and pass them onto each new root as it's created.

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -217,6 +217,7 @@ def _make_item(model):
     a ``.models.Document`` or ``.models.Resource``
     """
     item = Item(model.id, model.content, model.media_type)
+    return item
 
 
 def _node_to_model(tree_or_item, package, parent=None,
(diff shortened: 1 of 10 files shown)


reedstrm:
@karenc any comments? cte ruleset work is stuck behind this.

cnxepub/adapters.py
```diff
@@ -286,14 +286,19 @@ def __init__(self, item, package):
...
-        content = io.BytesIO(
-            b''.join([
-                isinstance(n, str) and n.encode('utf-8') or etree.tostring(n)
-                for n in self._html.xpath(content_xpath, namespaces=nsmap)]))
+        body = self._html.xpath('//xhtml:body',
+                                namespaces=HTML_DOCUMENT_NAMESPACES)[0]
+        metadata_node = self._html.xpath(
+                                    "//xhtml:body/*[@data-type='metadata']",
+                                    namespaces=HTML_DOCUMENT_NAMESPACES)
+        if len(metadata_node) > 0:
```
karenc: I wonder if it's better to do:

```python
for node in metadata_node:
    body.remove(node)
```

cnxepub/formatters.py
```diff
@@ -307,7 +317,7 @@ def __bytes__(self):
   </body>
 </html>
 """
-HTML_DOCUMENT = """\
+HTML_DOCUMENT = u"""\
```
karenc: This shouldn't be necessary because we had `from __future__ import unicode_literals`.

cnxepub/models.py
```diff
@@ -185,7 +196,7 @@ def flatten_to(model, flatten_filter):
     for m in flatten_model(model):
         if flatten_filter(m):
             yield m
-    raise StopIteration()
+    return
```
karenc: I see you're changing `raise StopIteration()` to `return`, maybe we don't even need `return`...?

karenc:
Do you have an example html that fails without this code change?  Just to be sure I correctly understand what you're trying to fix.

cnxepub/adapters.py
reedstrm: Hmm, I suppose there might be more than one metadata stanza, accidentally. That would be more robust. Will change.

cnxepub/formatters.py
reedstrm: Yeah, was getting a spurious error from the template formatter that googling indicated was am encoding issue. Was actually caused by interactions with the debugger. Will remove.

cnxepub/models.py
reedstrm: Python 3 complained about raise StopIteration() as deprecated, said an explicit return does that. I just tested, and in fact, implicit does as well. We can drop them completely.

reedstrm:
Yes, this is hard to test without the companion PR in rhaptos.cnxmlutils: https://github.com/Connexions/rhaptos.cnxmlutils/pull/145 That generates HTML with a `class` attrib on the  `body` tag, from CNXML that has such an attribute on the `document` tag. Hmm, you're right I didn't modify any of the existing tests to have that in place: I ran locally against a copy of stats that I generated. I'll add that feature to one of the existing HTML files.

reedstrm:
rebased to puck up the other PR, and squashed to hide my hassles getting the tests working correctly. :-)

PR #82, opened 2016-07-01
Namespace fixes

reedstrm:
conditional xpaths w/ and w/o xhtml namespace corrected.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -95,7 +95,7 @@ class HTMLFormatter(object):
             'blockquote', 'q', 'code', 'pre', 'object', 'img', 'audio',
             'video',
             ]
-        elements_xpath = '|'.join(['.//{}|xhtml:{}'.format(elem, elem)
+        elements_xpath = '|'.join(['//{}|//xhtml:{}'.format(elem, elem)
                                   for elem in elements])
 
         data_types = [
@@ -103,7 +103,7 @@ class HTMLFormatter(object):
             'footnote-number', 'footnote-ref', 'problem', 'solution', 'media',
             'proof', 'statement', 'commentary'
             ]
-        data_types_xpath = '|'.join(['.//*[@data-type="{}"]'.format(data_type)
+        data_types_xpath = '|'.join(['//*[@data-type="{}"]'.format(data_type)
                                      for data_type in data_types])
 
         xpath = '|'.join([elements_xpath, data_types_xpath])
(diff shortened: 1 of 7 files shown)


cnxepub/formatters.py
```diff
@@ -95,15 +95,15 @@ def _generate_ids(self, document, content):
             'blockquote', 'q', 'code', 'pre', 'object', 'img', 'audio',
             'video',
             ]
-        elements_xpath = '|'.join(['.//{}|xhtml:{}'.format(elem, elem)
+        elements_xpath = '|'.join(['//{}|//xhtml:{}'.format(elem, elem)
```
karenc: This change appears to generate ids for the top level `<div>`, but it's not reflected in the tests.  Are you sure this is correct?

cnxepub/tests/test_formatters.py
```diff
@@ -292,6 +292,7 @@ def test_document_auto_generate_ids(self):
         from ..formatters import HTMLFormatter
 
         random.seed(1)
+        n = random.randint(0, 100000)
```
karenc: This shouldn't be necessary.  If the test isn't passing without this, we need to investigate into it.

cnxepub/tests/test_formatters.py
```diff
@@ -433,6 +434,8 @@ def test_binder(self):
             expected_content = f.read()
 
         actual = str(SingleHTMLFormatter(self.desserts))
+        with open('/tmp/desserts-single-page-py2.xhtml', 'w') as f:
```
karenc: This should be removed.

cnxepub/formatters.py
karenc: The top level element is a temporary wrapper because etree requires a root element.  It gets removed when we change the content to string again.  I think we should use `.//`.

cnxepub/tests/test_formatters.py
reedstrm: yup missed that when re-cleaning.

reedstrm:
Clearly, I missed the import of the .// vs // in the xpaths, since I had it in my head that we only run these against the top level of any given Document. And usually, there's no difference, as long as your xpath doesn't match the root itself. Which in this case, it does.  will fix. As to the tree-w-root vs. string-HTML-fragment, I feel that's a model design decision that's going to need revisiting. I'm pretty sure that reverting the .// -> will also obviate the need for the uuid test changes. The main fix here was getting the non-namespaced|namespaced pairings correct.

karenc:
Comments are addressed in #85.

PR #83, opened 2016-07-01
rework adapt_single to have all page_ids available

reedstrm:
adapt_single_html_page recursed, so the code to check if a page was in this book was slightly off.


reedstrm:
fixed in #84

PR #84, opened 2016-07-01
only fixup refs for wholebook

reedstrm:
Alternate way to fix link fixup issue

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -349,7 +349,7 @@ def adapt_single_html(html):
     return binder
 
 
-def _adapt_single_html_tree(parent, elem, nav_tree):
+def _adapt_single_html_tree(parent, elem, nav_tree, pages_by_id=None, depth=0):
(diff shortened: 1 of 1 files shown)


reedstrm:
I'm going to merge this one for now. May need parts of the other approach if link fix up needs more tweaking, in the future.

cnxepub/adapters.py
```diff
@@ -418,3 +412,13 @@ def fix_generated_ids(page, content):
 
     for i, node in enumerate(parent):
         parent.set_title_for_node(node, title_overrides[i])
+
+    # For each page in the book, look for links to #auto_ and replace it with
+    # either the value without #auto_{id} or /contents/{id}#target
+    # only fixup links at top, for whole book
+    if depth == 0:
```
karenc: Are we only fixing the links if the page is directly under a book?  (So not for pages under chapters and units?)

karenc: Ah wait, I misread, is this to make sure the link fixing code only happen once?

PR #85, opened 2016-07-05
fix my breakage of id-adding xpaths and tests

reedstrm:
title say sit all. Making the xpaths global broke ID adding, which broke the tests in interesting ways. This reverts all the test changes, and makes the xapths relative again, while retaining the correct namespace-alternative behavior, that was the original point of the change.


karenc:
Looks great.  Thank you! :+1:

reedstrm:
Now that the test s have passed, feel free to merge :-)

PR #86, opened 2016-07-06
local intra-book link rewriting

reedstrm:
This converts links that look like /contents/<page-uuid>[#hash]  that link to other parts of the book, into local fragment links within the single-page html. This allows the target-link rules in the recipes to fill in the appropriate target-label value (including generated text labels and numbers, like Table 4.3)

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -9,6 +9,7 @@ from __future__ import unicode_literals
 import random
 import sys
 
+import re
 import jinja2
 import lxml.html
 from lxml import etree
@@ -16,6 +17,7 @@ from copy import deepcopy
 
 from .models import (
     model_to_tree,
+    flatten_to_documents,
     Binder, TranslucentBinder,
     Document, DocumentPointer, CompositeDocument, utf8)
 from .html_parsers import HTML_DOCUMENT_NAMESPACES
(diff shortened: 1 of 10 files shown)


karenc:
Is there a test to show that the links will be converted back to the correct url when adapting from single html?  (I haven't looked at the existing tests, it might already exist)

cnxepub/formatters.py
```diff
@@ -249,6 +251,20 @@ def _build_binder(self, binder, elem):
 
     def build(self):
         self._build_binder(self.binder, self.body)
+        # Rewrite absolute-path links that are intra-binder
+        page_ids = [page.id for page in flatten_to_documents(self.binder)]
+        page_uuids = {id.split('@')[0]: id for id in page_ids}
+        for link in self.root.xpath('//*[@href]'):
+            href = link.get('href')
+            if href.startswith('/contents/'):
```
philschatz: Could/should inter-module links be converted so their URL is `./{UUID}` instead of `/contents/{UUID}`? It seems like that would work for online, epub, hosting-yourself-by-unzipping-the-EPUB,  and it wouldn't hardcode `/contents/` into the paths if we switch to something like `/c/`

karenc: Let's create an issue for this.

karenc:
@reedstrm I added a commit for converting `"#page-id"` to `"/contents/page-id"` again when adapting from single html.  Please have a look when you have time.

PR #91, opened 2016-08-16
transparentBinder title may not have any text

reedstrm:
so let's use an HtmlElement and get all the child text content as well.

### DIFF
--- a/cnxepub/tests/test_collation.py
+++ b/cnxepub/tests/test_collation.py
@@ -288,38 +288,32 @@ body::after {
 
 /* copied from cte books/rulesets/common/toc.less */
 body > div[data-type="page"],
-body > div[data-type="composite-page"] {
-  pass: 20;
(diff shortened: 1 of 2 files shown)


reedstrm:
Seems previous marked-up-titles were passing validation at this point by virtue of having a single trailing space in the markup. Clearly, we don't use the title of the translucentBinder anywhere.

cnxepub/tests/test_collation.py
```diff
@@ -288,38 +288,32 @@ def test_with_ruleset(self):
 
 /* copied from cte books/rulesets/common/toc.less */
 body > div[data-type="page"],
-body > div[data-type="composite-page"] {
-  pass: 20;
+body > div[data-type="composite-page"]:pass(20) {
```
reedstrm: This was broken by change in cnx-easybake (pass: # is now :pass(3) )

PR #92, opened 2016-08-30
rewrite every href that resolves inside the book

reedstrm:
This will allow rulesets to create any `id` they want (as long as it's unique in the whole book) and create `href` links to them. These links will then be rewritten. Also gripes about bad `hrefs` (`#id` w/ no such `id` in the book)

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -347,35 +347,61 @@ def adapt_single_html(html):
     return binder
 
 
-def _adapt_single_html_tree(parent, elem, nav_tree, pages_by_id=None, depth=0):
+def _adapt_single_html_tree(parent, elem, nav_tree, id_map=None, depth=0):
     title_overrides = [i.get('title') for i in nav_tree['contents']]
 
-    def fix_generated_ids(page, content):
-        for i in content.xpath('.//*[starts-with(@id, "auto_")]'):
-            i.attrib['id'] = i.attrib['id'].split('_', 2)[-1]
-        for i in content.xpath('.//*[starts-with(@href, "#auto_")]',
+    # A dictionary to allow look up of a document and new id using the old html
+    # element id
+
+    if id_map is None:
+        id_map = {}
+
+    def fix_generated_ids(page, id_map):
+        """Fix element ids (remove auto marker) and populate id_map."""
+
+        content = content_to_etree(page.content)
+
+        new_ids = []
+        for element in content.xpath('.//*[@id]'):
+            id_val = element.get('id')
+            if id_val.startswith('auto_'):
+                new_val = id_val.split('_', 2)[-1]
+                # Did content from different pages w/ same original id
+                # get moved to the same page?
+                if new_val in new_ids:
+                    suffix = 0
+                    while (new_val + str(suffix)) in new_ids:
+                        suffix += 1
+                    new_val = new_val + str(suffix)
+            else:
+                new_val = id_val
+            new_ids.append(new_val)
+            element.set('id', new_val)
+            id_map['#{}'.format(id_val)] = (page, new_val)
+
+        id_map['#{}'.format(page.id)] = (page, '')
+        if page.id and '@' in page.id:
+            id_map['#{}'.format(page.id.split('@')[0])] = (page, '')
+
+        page.content = etree_to_content(content)
+
+    def fix_links(page, id_map):
+        """Remap all intra-book links, replace with value from id_map."""
+
+        content = content_to_etree(page.content)
+        for i in content.xpath('.//*[starts-with(@href, "#")]',
                                namespaces=HTML_DOCUMENT_NAMESPACES):
-            _, _, target = i.attrib['href'].split('_', 2)
-            target_page = pages_by_id[i.attrib['href']]
-            if target_page is page:
+            ref_val = i.attrib['href']
+            target_page, target = id_map[ref_val]
+            target_id = target_page.id.split('@')[0]
+            if page == target_page:
                 i.attrib['href'] = '#{}'.format(target)
+            elif not target:  # link to page
+                i.attrib['href'] = '/contents/{}'.format(target_id)
             else:
                 i.attrib['href'] = '/contents/{}#{}'.format(
-                    target_page.id, target)
-        target_pages = set(pages_by_id.values())
-        if target_pages:
-            page_ids_wo_versions = [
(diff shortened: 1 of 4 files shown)


pumazi:
Any chance we can get an adjustment to an existing test to make sure this line falls within those tests? https://coveralls.io/builds/7673742/source?filename=cnxepub%2Fadapters.py#L375

reedstrm:
Hmm, the single_collated_html test looks like it should test that, since it contains `href="#id-on-other-page" links. Hrmm.

reedstrm:
found an issue, hold one

cnxepub/adapters.py
```diff
@@ -347,35 +347,61 @@ def adapt_single_html(html):
...
+    def fix_generated_ids(page, id_map):
+        """Fix element ids (remove auto marker) and populate id_map."""
+
+        content = content_to_etree(page.content)
+
+        for element in content.xpath('.//*[@id]'):
+            new_ids = []
+            id_val = element.get('id')
+            if id_val.startswith('auto_'):
+                new_val = id_val.split('_', 2)[-1]
```
pumazi: Could you add an example of what this is splitting? I'm guessing something like `auto_generated_1234`?

cnxepub/adapters.py
```diff
@@ -347,35 +347,61 @@ def adapt_single_html(html):
...
+                # Did content from different pages w/ same original id
+                # get moved to the same page?
+                if new_val in new_ids:
```
pumazi: The guts of this conditional block aren't getting hit by any tests. Can you add a test for this?

reedstrm: Yup, I noticed, wrote a test, found a bug, and pushed it. (new_ids should be per page, not per element)

reedstrm:
@pumazi @karenc any further thoughts?

pumazi:
@reedstrm Just a word on the separate pep8 commit for future reference... One way to prevent the need for a separate commit would be to use `commit --amend` or later squash the commit into the previous commit while doing an interactive rebase `rebase -i <branch>`.

Another way is to use a fixup commit `commit --fixup=<commit-hash>` which allows you to fixup a past commit, several commits ago even, then use `rebase -i --autosquash <branch>`. This way is particularly good, because it's something you can do to someone else's commit, but at the same time see the change. Then when you are ready to merge those changes it becomes super easy to automatically squash them.

reedstrm:
normally I don't need a separate pep8 commit ever since I integrated flake8 into my editor on every save. But apparently I broke that in this virtual env, somehow. Fixed now. I'll squash.

reedstrm:
squashed, code identical to last run. I'll have to play more with `--fixup` and `autosquash`. I saw that they exists, but haven't dug into them.

cnxepub/adapters.py
```diff
@@ -347,35 +347,61 @@ def adapt_single_html(html):
...
+                        suffix += 1
+                    new_val = new_val + str(suffix)
+            else:
+                new_val = id_val
+            new_ids.append(new_val)
+            element.set('id', new_val)
+            id_map['#{}'.format(id_val)] = (page, new_val)
+
+        id_map['#{}'.format(page.id)] = (page, '')
+        if page.id and '@' in page.id:
```
karenc: Can `page.id` be empty?  If so, what happens to the line above this one? (line 382)

cnxepub/tests/test_adapters.py
```diff
@@ -824,6 +824,10 @@ def test_fix_generated_ids_in_composite_page(self):
                       extra.content)
         self.assertIn('Click <a href="/contents/extra#1">here</a>',
                       chocolate.content)
+        self.assertIn('<p id="summary0"> Pretend move of lemon summary</p>',
```
karenc: Are the links correct?  Is it going to link to `/contents/extra#summary0`?

cnxepub/adapters.py
reedstrm: One of the tests constructs a document w/ an empty id, so I put his guard in. I agree that it's unlikely to be a real case. The map gets an entry for '#None', which seems to be harmless.

cnxepub/tests/test_adapters.py
reedstrm: There are no links to that item. That would be a good test, I agree.

cnxepub/adapters.py
karenc: Oh I see. :+1:

cnxepub/adapters.py
```diff
@@ -347,35 +347,61 @@ def adapt_single_html(html):
...
-            _, _, target = i.attrib['href'].split('_', 2)
-            target_page = pages_by_id[i.attrib['href']]
-            if target_page is page:
+            ref_val = i.attrib['href']
+            target_page, target = id_map[ref_val]
+            target_id = target_page.id.split('@')[0]
+            if page == target_page:
                 i.attrib['href'] = '#{}'.format(target)
+            elif not target:  # link to page
+                i.attrib['href'] = '/contents/{}'.format(target_id)
```
philschatz: ooh, maybe out-of-scope but why is this `/contents/` instead of `./{}` or `../contents/{}`? Is it necessary to begin with a `/`? (Also, I _think_ this is what is causing links to break if you just do `wget -m https://archive/contents/{BOOK_UUID}`

philschatz:
Are there examples of how the HTML changes as a result of this code change? I'm curious why the string-detecting (`^auto_`) + splitting + concatenation is necessary to generate a new id (as opposed to just `id="auto_{ids++}"`)

reedstrm:
The 'auto_' etc. id's should never appear in actual content - they are an artifact of taking individual files, concatenating to a single-html, and reconstituting the separate files. As to syntax for 'link to other page', host-relative links is what we've been doing forever. I believe they work inside an epub as well, if you have a proper epub reader.

reedstrm:
also, where'd you get post-increment? No such beast in python ...

reedstrm:
added a link-and-test to the duplicate ids. used fixup and autosquash. v.nice.

cnxepub/adapters.py
reedstrm: `wget -m -k https://archive.cnx.org/contents/{BOOK_UUID}.html`

cnxepub/tests/test_adapters.py
reedstrm: added link and test

PR #94, opened 2016-09-22
Namespaces

reedstrm:
This code runs extra parsing steps on output of HTML from EPUB, to cleanup and regularize the namespaces. The output HTML is semantically identical to the previous version - the changes are purely syntax, making for a more human-readable output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 sudo: false
 python:
   - "2.7"
-  - "3.3"
+  - "3.5"
 before_install:
   # Install the coverage utility and coveralls reporting utility
   - pip install coverage
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -191,7 +191,7 @@ class HTMLFormatter(object):
 
     def __bytes__(self):
         html = self._template.render(self._template_args)
-        return html.encode('utf-8')
+        return _fix_namespaces(html.encode('utf-8'))
 
 
 class SingleHTMLFormatter(object):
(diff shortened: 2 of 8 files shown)


reedstrm:
BTW, given that the output is semantically identical, but  had to fix expected output anyway, means that our tests are too picky about exactly what constitutes a pass.

philschatz:
would `xmllint --format 2` help? (is there a python lib that wraps it?) I think it alphabetizes the attributes and normalizes the whitespace between attributes inside an element as well as whitespace in `text()` nodes.

reedstrm:
no, but `xmllint --c14n` does generate two files that differ only in that ElementTree strips unused namespace declarations.

reedstrm:
And is directly accessible from `lxml.etree` as `write_c14n()` method. Hmm.

pumazi:
Does this mean the content in archive will need to be need updated to contain all the namespaces?

pumazi:
I'm not really sure this extends to publishing yet, but if I'm not mistaken this should help us strip the unused namespaces that end up in our content as well as in the summary. Yeah?

reedstrm:
Yes, the c14n ("canonicalization") support may be the way to go for what we store.  I think I can handle the squash. I'll get on that now. The protocol now is since I've had your signoff, once I make a clean-looking set of commits, I can just merge it, since it'll be byte identical w/ this one, right?

reedstrm:
squashed, and `--force-with-lease` pushed. :-)

reedstrm:
I think existing content already has all its namespaces, or it wouldn't be well formed (though I'm not sure we test that on the (X)HTML ...

karenc:
I think it'll be great if we have tests passing at every commit. :D

karenc:
For example, when you change some code, include the changes for the tests in the same commit and avoid commits like "Python 3 compatibility". :D

reedstrm:
I think there's a line somewhere between "code commits exactly as I developed it" and the platonic ideal of a series of commits that each make a single atomic change, with tests, leading inevitably to the new feature implementation. I thought the division of code from test result changes in this case was warranted by the extent of the changes to the result sets, that just muddied the diff, and gave me a place to mention the reasons. In particular, archive tests take too long to run repeatedly during the heat of development. I'm willing to make some changes in how I work, but restructuring a set of commits beyond what's needed to make them more understandable feels like wasted time to me. I'm done with this code, I have other code to fix - if the commit history isn't misleading, I see little point in making it prettier.  The only easy thing I could have done in this particular case, AFAICT, is squash it all into one commit.

karenc:
There are 2 reasons I ask for commits like that: one is so we can understand the reason why something is changed, the other one is so when something goes wrong, we can try to find out exactly which commit broke what. If there's an intermediate commit that doesn't pass the tests, it is harder to do that kind of debugging. I think if your PR isn't that big and you can't be bothered to run tests, squashing all your commits into one is a good solution.

reedstrm:
BTW, that may have come across harsher than it should have. There's some truth that re-organizing the code so another dev can understand it helps you be sure you've actually covered all the corner cases, etc. Sort of like how you never know a subject until you teach it.

reedstrm:
Yup, git bisection is actually the best argument I've heard yet for a clean master commit history. THanks for reminding me of it.

reedstrm:
Squashed, with all the explication in the one commit message. This really is a pretty localized change, so shouldn't need bisecting

PR #95, opened 2016-10-13
Suppress empty summary

reedstrm:
Refs https://github.com/Connexions/webview/pull/1460

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,11 @@
+[flake8]
+exclude = 
+    .git,
+    .eggs,
+    cnx_epub.egg-info,
+    build,
+    dist
+
+putty-ignore =
+    cnxepub/tests/ : E501, E123, E126, F841
+    */__init__.py : F401, F403
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -611,12 +611,15 @@ HTML_DOCUMENT = """\
         </p>
       </div>
 
+      {% if metadata['summary'] %}
       <div class="description"
            itemprop="description"
            data-type="description"
            >
         {{ metadata['summary'] }}
       </div>
+      {%- endif %}
+
 
       {% for keyword in metadata['keywords'] -%}
       <div itemprop="keywords" data-type="keyword">{{ keyword|escape }}</div>
--- a/cnxepub/tests/test_formatters.py
+++ b/cnxepub/tests/test_formatters.py
@@ -433,6 +433,10 @@ class SingleHTMLFormatterTestCase(unittest.TestCase):
             expected_content = f.read()
 
         actual = str(SingleHTMLFormatter(self.desserts))
+        # Uncomment this if/when you get a test failure
+        # with open(os.path.join(TEST_DATA_DIR,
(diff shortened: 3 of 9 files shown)


cnxepub/tests/test_formatters.py
```diff
@@ -433,6 +433,10 @@ def test_binder(self):
             expected_content = f.read()
 
         actual = str(SingleHTMLFormatter(self.desserts))
+        # Uncomment this if/when you get a test failure
```
karenc: Hm... Not sure if it's a good idea.  It's not that hard to add this if necessary is it?

cnxepub/tests/data/test.xhtml
```diff
@@ -0,0 +1,314 @@
+<html xmlns="http://www.w3.org/1999/xhtml">
```
reedstrm: oops. Accidental `git add`. Will remove.

cnxepub/tests/test_formatters.py
reedstrm: This is the 4th time I've typed that from memory. Usually get it wrong at least once. :-/

.flake8
```diff
@@ -0,0 +1,12 @@
+[flake8]
+exclude = 
+#    tests,
```
reedstrm: yeah, handled by the next stanza down, where I use flake8-putty to ignore a subset of errors (mostly line length, and some style things about unused variables)

reedstrm: Grr, I made the fixes Friday, even committed, but failed to push! :facepalm:

cnxepub/tests/test_formatters.py
reedstrm: Hmm, that works, but makes the test less efficient.  One of the other tests does do that, over in archive, I think). o.k. let me see what I can cook up. Ohh, can I make it conditional on the test passing? Then there's no commenting to do!

karenc: Looks much better :smile: . Why not use a fixup commit?

cnxepub/tests/test_formatters.py
```diff
@@ -433,7 +433,16 @@ def test_binder(self):
             expected_content = f.read()
 
         actual = str(SingleHTMLFormatter(self.desserts))
+        out_path = os.path.join(TEST_DATA_DIR,
+                                'desserts-single-page-actual.xhtml')
+        if not IS_PY3:
+            out_path = out_path.replace('.xhtml', '-py2.xhtml')
+
+        with open(out_path, 'w') as out:
+                out.write(actual)
```
pumazi: wild indent here...

cnxepub/formatters.py
```diff
@@ -604,12 +604,15 @@ def _fix_namespaces(html):
...
 
+      {% if metadata['summary'] %}
       <div class="description"
            itemprop="description"
            data-type="description"
            >
         {{ metadata['summary'] }}
       </div>
+      {%- endif %}
+
```
pumazi: extra line...

reedstrm: not extra I'm pretty sure the tests don't pass without it.

PR #96, opened 2016-10-17
implement includes callback

reedstrm:
In support of pulling exercises from a remote server into the raw HTML, so that it is present at cooking, this code extends the SingleHTMLFormatter to take a list of tuples, as `(xpath, callback)` pairs. When the single HTML page has been built, these callbacks are called, once with each element that matches the `xpath`. Test example matches all `a` elements, and upcases the link text.

Next step is for publish to build a callback for inserting exercises, based on config values.

### DIFF
--- a/cnxepub/collation.py
+++ b/cnxepub/collation.py
@@ -51,21 +51,19 @@ def reconstitute(html):
     return adapt_single_html(xhtml)
 
 
-def collate(binder):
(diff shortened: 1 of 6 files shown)


reedstrm:
Oh, and one "while I'm in here" change: since I changed the signature for `collate`, make the name of the ruleset file to use for collating an optional parameter, defaulting to `ruleset.css`

PR #97, opened 2016-10-19
Include callback

reedstrm:
After testing with real data, found a couple cases to deal with. Resolving those required changing to XML formatted output, which changed some pretty-printing in the expected results.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -286,7 +286,6 @@ class SingleHTMLFormatter(object):
             self.build()
         return _fix_namespaces(etree.tostring(self.root,
                                               pretty_print=True,
-                                              method='html',
                                               encoding='utf-8'))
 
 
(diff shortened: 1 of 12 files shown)


reedstrm:
Yay! fixed coverage!

setup.py
```diff
@@ -9,6 +9,7 @@
 install_requires = [
     'jinja2',
     'lxml',
+    'requests',
```
pumazi: This is likely a `collation_requires` value, but it won't hurt to include it since most other applications use it anyhow.

cnxepub/formatters.py
```diff
@@ -319,9 +318,63 @@ def _fix_namespaces(html):
...
+def exercise_callback_factory(match, url_template):
+    """Create a callback function to replace an exercise by fetching from
+    a server."""
+    import requests
+
+    def _replace_exercises(elem):
+        item_code = elem.get('href')[len(match):]
+        url = url_template.format(itemCode=item_code)
+        res = requests.get(url)
+        if res:
```
philschatz: Could/should this error instead of silently failing? Otherwise, it seems like people would incorrectly think that a book is successfully generated.

PR #101, opened 2016-10-27
Exercise includes extensions - tokens, and math conversion

reedstrm:
Allow use of a token for authenticated access to get answers. Exact details of how to get the token, and when the service will be given a token, are policy issues to be handled upstream of this library.
Detect TeX formatted math in exercise HTML and convert it in-place to mathml, via mathmlcloud service.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -317,7 +317,7 @@ def _fix_namespaces(html):
     return etree.tostring(let, pretty_print=True, encoding='utf-8')
 
 
-def exercise_callback_factory(match, url_template):
+def exercise_callback_factory(match, url_template, token=None):
     """Create a callback function to replace an exercise by fetching from
     a server."""
     import requests
@@ -325,7 +325,11 @@ def exercise_callback_factory(match, url_template):
     def _replace_exercises(elem):
         item_code = elem.get('href')[len(match):]
         url = url_template.format(itemCode=item_code)
-        res = requests.get(url)
+        if token:
+            headers = {'Authorization': 'Bearer {}'.format(token)}
+            res = requests.get(url, headers=headers)
+        else:
+            res = requests.get(url)
         if res:
             # grab the json exercise, run it through Jinja2 template,
             # replace element w/ it
@@ -365,7 +369,9 @@ EXERCISE_TEMPLATE = jinja2.Template("""\
             {% if question.answers %}
             <ol data-number-style="lower-alpha">
                 {% for answer in question.answers %}
-                    <li>{{ answer.content_html }}</li>
+                    <li{% if 'correctness' in answer
+                        %} data-correctness={{ answer.correctness }}{%
+                    endif %}>{{ answer.content_html }}</li>
                 {% endfor %}
             </ol>
             {% endif %}
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -15,6 +15,8 @@ import lxml.html
 from lxml import etree
 from copy import deepcopy
 
+import requests
+
 from .models import (
     model_to_tree,
     flatten_to_documents,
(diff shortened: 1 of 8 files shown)


reedstrm:
note: the failure to pass on python2 seems to be due to libxslt-dev 1.2.26 (travis) vs 1.2.28 (my local install and the servers). Coverage is just wrong: it shows a decrease in coverage in the models.py file, but I changed usage of models not at all. The added lines in formatters.py is also wrong: while I did add 16 lines, only 1 of them is not covered (returning a None). Ah, I bet coverage is screwed up by the python2 tests not completing - the diff may be correct in that case.

cnxepub/tests/test_formatters.py
```diff
@@ -10,25 +10,124 @@
 import os
 import sys
 import unittest
+from lxml import etree
```
pumazi: FYI, not a big deal, but the import grouping here was correct... For future reference, things are typically grouped by 1) standard library 2) third-party 3) local.

cnxepub/tests/test_formatters.py
```diff
@@ -112,22 +211,57 @@ def last_extension(*args, **kwargs):
...
     # modified from http://stackoverflow.com/a/28507806/5430
-    class MockResponse:
-        def __init__(self, json_data, status_code):
-            self.json_data = json_data
-            self.status_code = status_code
 
-        def json(self):
-            return self.json_data
+    if 'headers' in kwargs:
+        assert kwargs['headers'] == {'Authorization': 'Bearer somesortoftoken'}
```
pumazi: Is it necessary to do this assert twice? It's repeated below...

cnxepub/formatters.py
```diff
@@ -317,15 +319,36 @@ def _fix_namespaces(html):
...
+    if res:
+        eq = res.json()
+        mml = etree.fromstring(eq['components'][0]['source'])
+        if node.tag.endswith('span'):
+            mml.set('display', 'inline')
+        elif node.tag.endswith('div'):
+            mml.set('display', 'block')
+        return mml
+    else:
+        return None
```
pumazi: This is never hit in coverage, but I'm not sure it can be hit. `requests.post` will always return a response, so the logic won't ever get here.

cnxepub/formatters.py
```diff
@@ -317,15 +319,36 @@ def _fix_namespaces(html):
...
     def _replace_exercises(elem):
         item_code = elem.get('href')[len(match):]
         url = url_template.format(itemCode=item_code)
-        res = requests.get(url)
+        if token:
+            headers = {'Authorization': 'Bearer {}'.format(token)}
+            res = requests.get(url, headers=headers)
+        else:
+            res = requests.get(url)
         if res:
```
pumazi: Not part of this pull request, but again this is probably not necessary. I guess it's something I missed in a past PR. 😞

cnxepub/formatters.py
philschatz: If the code should never be executed should this line throw an error to make it clear that this code path is never evaluated?

reedstrm: requests  Response object is special if 'OK' (basically 2XX or 3XX): 
```
>>> res=requests.get('http://example.com/nothing')
>>> res
<Response [404]>
>>> if res:
...   print True
... else:
...   print False
... 
False
>>> 
```
Code is here: https://github.com/kennethreitz/requests/blob/master/requests/models.py#L622-L628

cnxepub/tests/test_formatters.py
reedstrm: Yeah, I accidentally moved that. (I duped it, then flake8 caught the double import, and I removed the wrong one) I'll fix.

cnxepub/tests/test_formatters.py
reedstrm: I split the mock, and failed on refactoring that. Will fix.

reedstrm:
Any thoughts on what to do about the test diffs due to differing libxslt1 versions? (I haven't actually proved that that is the problem, but it seems the only significant diff in the build)

pumazi:
On a separate branch I updated the travis config to use ubuntu 14.04 rather than 12.04. The same failure showed up.

pumazi:
Also, I have libxslt==1.1.29. And the tests pass. These error messages about unicode are new though.

reedstrm:
yikes, I don't see those locally

reedstrm:
changing the logger in the cli script  back to the script name has removed the unicode errors.

reedstrm:
I'm going to do a selective squash rebase on this.

reedstrm:
Done rebasing/squashing @pumazi got any more comments?

PR #105, opened 2016-12-06
add default for the mathmlcloud URL commandline argument

philschatz:
I had trouble finding the API URL (and port) for the currently-running MathMLCloud instance. It might be better to switch it to run on 443 (instead of 1337) but this at least documents it a little bit more.

@reedstrm remembered the URL by looking through the tests although it seems the tests use port 80

Also, I did not test this code, just cargo-culted to make sure the bits were not lost

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ before_install:
   # Install the coverage utility and coveralls reporting utility
   - pip install coverage
   - pip install coveralls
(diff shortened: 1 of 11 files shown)


reedstrm:
I could trivially add "always use the default if no switch", but am worried that there's a use case for not being network dependent for this.

PR #106, opened 2016-12-12
handle shortid as well as fullid in metadata

reedstrm:


philschatz:
What if the short id was in the OPF file? Something like (from http://www.idpf.org/epub/31/spec/epub-ocf.html ):

```xml
<package
         xml:lang="en"
         xmlns="http://www.idpf.org/2007/opf">
    <manifest>
        <item id="${SHORT_ID}" href="contents/${UUID}.html" 
              media-type="application/xhtml+xml"/>
        <item id="${SHA}" 
              href="resources/${SHA}.png" 
              media-type="image/png"/>
    </manifest>
</package>
```

I _think_ the id's are currently only used for the `<spine>` (which we do not use). Thoughts?

reedstrm:
Yup, when I get to making this a valid epub, I'll probably use the shortid there. However, I think it's important that the ID be directly inside the content. I made it a data-* element on an empty node my parallel w/ cnx-archive-uri. Not sure why that one is that way, either. :-) I guess the default render is empty, then.

reedstrm:
Not sure why travis python 3.5.2 won't pass the tests - they pass  locally for me with the exact same python version.

reedstrm:
BTW, re: valid epub and spine: our current manifest is invalid in a number of ways - we repeat things in the manifest that are required to be unique. There are currently no ids, which are required. And there's no nav element, which is also required.

philschatz:
hmm, I think browsers sometimes remove empty span tags which might be problematic. Also, we sometimes show metadata and sometimes hide it (in attributes) which seems like it could be more consistent. Just posting up thoughts to ponder/discuss 😄

reedstrm:
These HTML files are not intended for a browser directly, though I can say both Helene and I have opened them in browsers, and not seen anything disappear, so I'm calling red herring on that one. Agreed that we need a more consistent data design.

philschatz:
RE red herring: I thought TEA was planning on using these HTML files in the browser with just CSS changes, no? It seems like the list of intentions for this ePub is already growing. but yep, agree on consistency

reedstrm:
Ah, well, perhaps a pink herring. :-) They will use them in a browser, but parse them on the way in, so even if a browser were to drop a span, they've already pulled out the interesting metadata at import time.

reedstrm:
Coverage dropped because I disabled the python3 tests, so there are branches that are only tested under python3 (compatibility imports, mostly)

reedstrm:
pinned lxml version, reactivated python3. We're good to go!

philschatz:
Do the instructions (or `setup.py`) need to be updated to reflect the specific version of lxml needed?

reedstrm:
No, only travis seems to need 3.6.4. THe latest version (3.7.0) works fine locally.

PR #107, opened 2016-12-14
Copy resources

reedstrm:
This enables a workflow for creating the baked zip directly from the raw epub and the baked single-page html file. Note I went with the second parameter, since it's both more explicit, and probably more scriptable.

### DIFF
--- a/cnxepub/scripts/collated_single_html/main.py
+++ b/cnxepub/scripts/collated_single_html/main.py
@@ -42,13 +42,15 @@ def main(argv=None):
 
     args = parser.parse_args(argv)
 
+    if args.input and args.output == sys.stdout:
+        raise ValueError('Cannot output to stdout if reading resources')
+
     from cnxepub.collation import reconstitute
     binder = reconstitute(args.collated_html)
 
     if args.dump_tree:
         print(pformat(cnxepub.model_to_tree(binder)),
               file=sys.stdout)
-
     if args.output:
         cnxepub.adapters.make_epub(binder, args.output)
 
--- a/cnxepub/scripts/collated_single_html/main.py
+++ b/cnxepub/scripts/collated_single_html/main.py
@@ -14,8 +14,8 @@ import argparse
 import logging
 import sys
 from pprint import pformat
-from lxml import etree
 import cnxepub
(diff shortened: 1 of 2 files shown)


philschatz:
Will the existing code that uses this still work? [Example](https://github.com/Connexions/cnx-recipes/blob/master/scripts/bake-book#L91-L93)

reedstrm:
Yes, existing code will work fine, this just adds an option. Yes, a test would not be out of place, (what, you channeling Karen?) but I decided to take a crack at pdf exercises instead.

cnxepub/scripts/collated_single_html/main.py
```diff
@@ -34,9 +34,12 @@ def main(argv=None):
     parser.add_argument('-d', '--dump-tree', action='store_true',
                         help='Print out parsed model tree.')
 
-    parser.add_argument('-o', '--output', type=argparse.FileType('w'),
+    parser.add_argument('-o', '--output', type=argparse.FileType('w+'),
```
pumazi: FYI, I believe the filetype allows for output direction. For example, `cat foo.zip | $script -i - -o - > bar.zip`, but this will fail because the python zipfile module can't use a streaming file buffer. Do to this I've tried not to use `argparse.FileType` for input/output operations.

reedstrm: cnx-epub-validate-collated -i data/TEAapphysics2-raw.epub -o - data/TEAapphysics2-baked.html >foo2.epubI just tested it. It worked.

reedstrm: By that I mean foo2.epub contains a valid zip with all the expected contents.

reedstrm: And that's a lie. Turns out that it didn't extend properly, foo2 has the copied resources but _not_ the original content.

reedstrm: And now it does the right thing, and throws an error inside pack_epub, since the zipfile class can't scroll the stdout.

reedstrm: I added guard code for this case, now.

PR #108, opened 2017-01-06
make missing exercises explicit

reedstrm:
Don't just delete missing exercises, leave a marker. (Which can then be used by a ruleset to delete more structure, if needed)
TeX conversion:
  + Convert the `data-math` attribute contents by preference.
  + Log bad TeX conversion.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -397,8 +397,9 @@ def exercise_callback_factory(match, url_template, token=None, mml_url=None):
                             mparent = node.getparent()
                             mparent.replace(node, mathml)
                         else:
+                            mathtext = node.text or ''
                             logger.warning('BAD TEX CONVERSION: "%s" URL: %s'
-                                           % (node.text.encode('utf-8'), url))
+                                           % (mathtext.encode('utf-8'), url))
 
             parent = elem.getparent()
             if etree.QName(parent.tag).localname == 'p':
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -325,14 +325,18 @@ def _fix_namespaces(html):
     return etree.tostring(let, pretty_print=True, encoding='utf-8')
 
 
-def _replace_tex_math(node, mml_url, depth=0):
+def _replace_tex_math(node, mml_url, retry=0):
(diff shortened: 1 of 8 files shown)


philschatz:
Could/should the validation fail if `.missing-exercise` is found?

reedstrm:
Interesting idea, but that's an editorial decision - the ruleset could very well just delete them.

philschatz:
I was trying to understand the bath mathml conversion case; does it silently convert the `data-math` attribute to be `data-math="retry"`?

Also, would it make sense to have a `--strict` CLI that would fail on missing exercises/LaTeX-conversion?

reedstrm:
No, bad TeX conversion is left in place after two retries. The example I mocked is a string 'retry'. While the cli is named '*-validate' it's not really a validator _per se_. It currently mostly validates by throwing errors if you give it poorly structured content. I think we need a real validator layer that does things like ensure all the linked-to-images are present in the zip, etc.

reedstrm:
BTW, forcing myself to get the coverage up actually found errors in the error handling. :-/

philschatz:
oh, I didn't realize it was a "validator" (air quotes)...

Seeing path coverage as well as project coverage delta is really nice/motivating!

cnxepub/formatters.py
```diff
@@ -347,8 +351,8 @@ def _replace_tex_math(node, mml_url, depth=0):
         else:
             logger.warning('Retrying math TeX conversion: '
                            '{}'.format(json.dumps(eq, indent=4)))
-            if depth < 2:
-                return _replace_tex_math(node, mml_url, depth)
+            if retry < 2:
+                return _replace_tex_math(node, mml_url, retry)
```
pumazi: Perhaps a loop should be used rather than recursion. This puts an unnecessary parameter on the function.

cnxepub/formatters.py
```diff
@@ -386,6 +396,10 @@ def _replace_exercises(elem):
                         if mathml is not None:
                             mparent = node.getparent()
                             mparent.replace(node, mathml)
+                        else:
+                            mathtext = node.get('data-math') or node.text or ''
```
pumazi: The `... or ... or ...` is probably not the most legible way to write this.

PR #109, opened 2017-01-30
IDs for subcollections (uuids and shortIds)

reedstrm:
add shortId to all models - optionally allow container objects to have an Id other than 'subcol' (use child Binder instead of TransluncentBinder for those cases)

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -229,8 +229,15 @@ def _node_to_model(tree_or_item, package, parent=None,
         if tree['id'] == lucent_id:
             binder = TranslucentBinder(metadata={'title': tree['title']})
         else:
-            package_item = package.grab_by_name(tree['id'])
-            binder = BinderItem(package_item, package)
+            try:
+                package_item = package.grab_by_name(tree['id'])
+                binder = BinderItem(package_item, package)
+            except KeyError:  # Translucent w/ id
+                binder = Binder(tree['id'],
+                                metadata={
+                                   'title': tree['title'],
+                                   'cnx-archive-uri': tree['id'],
+                                   'cnx-archive-shortid': tree['shortId']})
         for item in tree['contents']:
             node = _node_to_model(item, package, parent=binder,
                                   lucent_id=lucent_id)
@@ -415,7 +422,15 @@ def _adapt_single_html_tree(parent, elem, nav_tree, id_map=None, depth=0):
                         child.xpath('*[@data-type="document-title"]',
                                     namespaces=HTML_DOCUMENT_NAMESPACES)[0]
                         ).text_content().strip()
-            tbinder = TranslucentBinder(metadata={'title': title})
+            metadata = parse_metadata(child)
+            id_ = metadata.get('cnx-archive-uri') or child.attrib.get('id')
+            shortid = metadata.get('cnx-archive-shortid')
+            if id_ is None:
+                tbinder = TranslucentBinder(metadata={'title': title})
+            else:
+                tbinder = Binder(id_, metadata={'title': title,
+                                                'id': id_,
+                                                'shortId': shortid})
             _adapt_single_html_tree(tbinder, child,
                                     nav_tree['contents'].pop(0),
                                     id_map=id_map, depth=depth+1)
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -786,9 +786,20 @@ HTML_DOCUMENT = """\
 #      a way to render the the tree to html. This either needs to
 #      move elsewhere or preferably be replaced with a better solution.
 def html_listify(tree, root_xl_element, extensions, list_type='ol'):
+    """Convert a node tree into an xhtml nested list-of-lists.
+
+       This will create 'li' elements under the root_xl_element,
+       additional sublists of the type passed as list_type. The contents
+       of each li depends on the extensions dictonary: the keys of this
+       dictionary are the ids of tree elements that are repesented by files
+       in the epub, with associated filename extensions as the value. Those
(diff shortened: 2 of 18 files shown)


reedstrm:
THis needs support in single-html and validate-collated, yet.

cnxepub/formatters.py
```diff
@@ -775,9 +775,13 @@ def _replace_exercises(elem):
 def html_listify(tree, root_xl_element, extensions, list_type='ol'):
```
pumazi: Any chance you could throw a quick docstring up here? I seem to have lost track of what `extensions` is...

cnxepub/models.py
```diff
@@ -128,12 +128,12 @@ def model_to_tree(model, title=None, lucent_id=TRANSLUCENT_BINDER_ID):
         {'id': <id>|'subcol', 'title': <title>, 'contents': [<tree>, ...]}
 
     """
-    if type(model) is TranslucentBinder:
+    id = model.ident_hash
+    if id is None and isinstance(model, TranslucentBinder):
```
pumazi: Given a non-identified binder (i.e. `binder.id is None`), the resulting id in the tree would be `subcol`. `type(model)` was used here because we want to know the exact type of the instance. The evaluation for this looks something like:
```python
assert isinstance(binder, TranslucentBinder)  # ✅ 
assert isinstance(translucent_binder, TranslucentBinder)  # ✅ 
assert isinstance(document, TranslucentBinder)  # ❌ 
```

reedstrm: eventually, all binders and sub-binders will have Ids so, this won't matter. ARe you suggesting we should allow unidentified binders out to pass, with id=None ?

cnxepub/formatters.py
reedstrm: Yeah I had to dig for that, too - it's the extension to use for the filename for this object. It's "xhtml" for almost everything in our epubs. I'm using this as a handy way to detect the objects that don't have a file: i.e. subcollections. I'll add some comments

cnxepub/models.py
pumazi: The cnx-epub library does not enforce identifiers. It should be possible to build unidentified content object structures. The library itself isn't concerned about identifiers or what the identifiers look like. This is why things like `cnx-archive-uri` exist. Technically, it should be possible to incorporate another provider's identifier scheme into this library.

Unfortunately I don't feel like everyone is sharing this vision of a library that is identifier and persistence layer agnostic. So I may decide to roll over on the subject. 

I'm feeling like the vision is already quite broken. And it will be increasingly broken with the incorporation of the external identifier scheme (i.e. `shortId`).

reedstrm: Hate to see a vision die, but to be honest, it's never been that clear. We waffled a lot about "zipped up, epub-like-not-really-epub-format" at the beginning. And cnx-epub does not in fact handle general pubs, following the spec: it doesn't create a spec compliant opf. It says in the README "Library for parsing and building EPUB3 files for connexions content." And we need a central library to enforce *Connexions* specific content limitations - i.e. presence of a license, at least one author, or the document having a unique ID. If cnx-epub isn't the place for that, where is? Not archive, I don't think - it's domain is persisting and serving the content. Ideally, it should know as little as possible about the internal structure. Perhaps publish.
I think there's probably a need for both - a generic epub handler, and code that enforces the Connexions/OpenStax specialization of that. I have a feeling cnx-epub might grow(split?) into that, as we put more  requirements in place  - creating user-facing epubs, importing arbitrary (broken?) epubs, dealing with epub editors. For the future, I'll try to see the overall shape of the code a bit better, see if we can nudge it that way.

P.S. this change (adding ids and metadata to a level of the data structure that did not have it before) is not minor surgery. I'm actually well pleased with how _little_ code changes are needed, mostly bypassing exceptions that had been coded in for the `subcol` special case. That speaks to the flexibility of the original design.

reedstrm: re: the specific code, I know how it evaluates - binders are a subclass of translucent binders. The previous code expected TLs to not have an id, so create one here so that the tree does not have id=None nodes for that case. It would have left a top level node for a putative un-identifed book as that id=None. If we're going to support one, why not the other? Also, passing "None" around feels wrong to me - bit of a code smell - see all the cleanup in the around required metadata in the tests. IT feels like mis-labelling a non-id'ed book as a subcol may be the lesser sin.

reedstrm: In fact, the docstring for that function implies that if you have no id, you'll get the id 'subcol'

reedstrm:
dealing with coverage now. It's legit, I added paths w/o adding examples of that type to the test data.

reedstrm:
re: coverage of diff - I touched a line that uses a random nmber, and is only hit if the randomly chosen replacement ID happens to hit an existing ID. It's still not covered, and hacking up  way to cover it is not top of my agenda today.

reedstrm:
@pumazi I'm going to stop making changes now. I think I've addressed your review above.

cnxepub/adapters.py
```diff
@@ -229,8 +229,15 @@ def _node_to_model(tree_or_item, package, parent=None,
         if tree['id'] == lucent_id:
             binder = TranslucentBinder(metadata={'title': tree['title']})
         else:
-            package_item = package.grab_by_name(tree['id'])
-            binder = BinderItem(package_item, package)
+            try:
+                package_item = package.grab_by_name(tree['id'])
+                binder = BinderItem(package_item, package)
+            except KeyError:  # Translucent w/ id
+                binder = Binder(tree['id'],
```
pumazi: This breaks with the idea that this is an adaptation not a conversion.

reedstrm: Ah, BinderItem vs. Binder.

PR #111, opened 2017-02-23
disallow @version in model ids, split them to id and metadata["version"]

reedstrm:
This is the last step for stabilzing the assigned uuids for composite documents - we removed assigning uuids from rulesets/easybake in order to do it here when publishing. Not yet complete, needs testing in the publish-new-version case, to verify that uuids stay constant, and the versions bump appropriately.

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -6,14 +6,18 @@
 # See LICENCE.txt for details.
 # ###
 from __future__ import unicode_literals
-import os
+import base64
 import io
 import logging
 import mimetypes
+import os
+import uuid
+
 from copy import deepcopy
 
 import jinja2
 import lxml.html
+
 from lxml import etree
 
 from .epub import EPUB, Package, Item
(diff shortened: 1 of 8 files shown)


reedstrm:
More steps making this less general, more cnx specific. We need to be able to do something like this, though. Also, I see nothing in here when publishing composite modules that will increment versions, so it's likely there's still lurking bugs.

reedstrm:
Ctaully, the version problem is solved by using the parent version - ignore that. PErhaps this _is_ ready to merge, at least to get it on dev.cnx.org so we can do more tests there. Need to get this to tea.cnx.org, so we can use it to generate the zip/epub deliverables from that content.

reedstrm:
I'll update the tests to get the coverage up to snuff.

reedstrm:
yay! code coverage wins, again. Had 'protect exceptional cases' fixes to both functions that had not been covered in the original tests.

PR #118, opened 2017-05-09
Exercise fetch - caching and missing handing

reedstrm:
Use a memcache client, if provided, in the callback function. Provide commandline option to `cnx-epub-single-html` to provide a memcache server, defaulting to localhost:11211.


reedstrm:
I need to add tests to this that use the memcache - I'll copy from cnx-archive

Issue #117, opened 2017-05-05
metadata parsing should have different magic


reedstrm:
Currently, the MetaDataParser contains magic attributes, like this one:
https://github.com/Connexions/cnx-epub/blob/master/cnxepub/html_parsers.py#L200
The effectively search _down_ the HTML tree for values. These should search _up_  the tree, using something like this (e.g. for license_url):
node.xpath('*[@data-type="metadata"]//*[@data-type="license"]/@href | ancestor::*/*[@data-type="metadata"]//*[@data-type="license"]/@href')[0]

That will find the value from the metadata child, or the metadata child of any ancestor, so for page it would inherit the chapter, or book. For a chapter, it would inherit from the book.

PR #119, opened 2017-06-26
use HTML parsing for baking. Make html_parsers namespace agnostic

reedstrm:
The cnx-easybake cli tool uses the lxml.etree.HTMLParser. Internal cnx-epub baking did not: it used the XML parser. This came to light when sorting of keywords broke. This occured because the CSS selector used in recipes to designate the text to sort on (e.g. `dl > dt`) is converted internally by easybake to an xpath: `descendant-or-self::dl/dt` in order to create a sorting function to apply to each node when doing the sorted output. This would not match namespaced nodes. Ideally, this code should instead build another `cssselect2.Match` object, instead, just like the main processing loop in easybake. However, I would then need to create the equivalent to `etree.tostring`,  and other machinery. I started down that path, but believe I've ran into incomplete parts of `cssselect2`, and potentially just outright bugs. (the `query` function of an `ElementWrapper` claims to take `CompiledSelectors`, but errors when I try that) I will probably need to do that eventually, but this seemed a less invasive immediate fix.

Note that most of the xpath changes are removing unneeded namespace declarations. While this potentially changes the output in a mixed-xml environment (say some SVG or mathml) I don't think the results would be suprising.  There are a number of other simple alternative xpaths (with and w/o namespace, i.e. `xhtml:a | a`).

The only complicated cases are `license_url` and `license_text`. I in fact altered the behavior those two metadata methods to handle the now better understood business logic: any element should be licensed under its directly declared license, or that of one of its ancestors. I also added the special case of checking for a specific top level license for the whole book - this doesn't fall under the other rule if you happen to be holding the `html` or `head` nodes. Note that only this last case ends up needing the namespace alternatives.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -539,6 +539,7 @@ HTML_DOCUMENT = """\
       xmlns:mod="http://cnx.rice.edu/#moduleIds"
       xmlns:md="http://cnx.rice.edu/mdml"
       xmlns:c="http://cnx.rice.edu/cnxml"
+      lang="{{ metadata['language'] }}"
       >
   <head itemscope="itemscope"
         itemtype="http://schema.org/Book"
--- a/cnxepub/html_parsers.py
+++ b/cnxepub/html_parsers.py
@@ -65,7 +65,7 @@ def _nav_to_tree(root):
             shortid = li.get('cnx-archive-shortid')
             yield {'id': itemid,
                    # Title is wrapped in a span, div or some other element...
-                   'title': _squash_to_text(expath(li, 'xhtml:*')[0],
+                   'title': _squash_to_text(expath(li, '*')[0],
                                             remove_namespaces=True),
                    'shortId': shortid,
                    'contents': [x for x in _nav_to_tree(li)],
@@ -86,7 +86,7 @@ def parse_metadata(html):
 
 def parse_resources(html):
     """Return a list of resource names found in the html metadata section."""
-    xpath = '//xhtml:*[@data-type="resources"]//xhtml:li/xhtml:a'
+    xpath = '//*[@data-type="resources"]//xhtml:li/xhtml:a'
     for resource in html.xpath(xpath, namespaces=HTML_DOCUMENT_NAMESPACES):
         yield {
             'id': resource.get('href'),
@@ -141,7 +141,7 @@ class DocumentMetadataParser:
 
     @property
     def title(self):
-        items = self.parse('.//xhtml:*[@data-type="document-title"]/text()')
+        items = self.parse('.//*[@data-type="document-title"]/text()')
         try:
             value = items[0]
         except IndexError:
@@ -150,7 +150,7 @@ class DocumentMetadataParser:
 
     @property
     def summary(self):
-        items = self.parse('.//xhtml:*[@data-type="description"]')
+        items = self.parse('.//*[@data-type="description"]')
         try:
             description = items[0]
             value = _squash_to_text(description).encode('utf-8')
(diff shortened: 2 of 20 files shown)


reedstrm:
This impacts cnx-recipes#179 as well as #117

cnxepub/collation.py
```diff
@@ -33,7 +33,8 @@ def easybake(ruleset, in_html, out_html):
     with respective read and write ability.
 
     """
-    html = etree.parse(in_html)
+    html_parser = etree.HTMLParser(encoding="utf-8")
```
philschatz: I missed the discussion, what was the reason for using the HTML parser? I didn't see that part explained in the PR description

reedstrm: Was the PR description tl;dr? It's nothing but a description of the problem and why this workaround was chosen over rewriting the selector parser in easybake.

philschatz: ah, easybake uses the HTML parser because the CSS would need to use the XHTML namespaces (via `@namespace 'http://www.w3.org/1999/xhtml`;` at the top of the CSS). Gotcha

reedstrm: Umm, @philschatz , I respect your coding chops, but I doubt it's a quick fix. The extra gnarly bit is that the `sort-by/group-by` code isn't even using `cssselect2` right now, but `cssselect`'s `xpath` functionality. That needs to be converted over as well.

reedstrm: And that's where the bug shows up - the sorting. (Actually, now that I think about it, I'm sort of surprised _any_ of the selectors match …)

pumazi:
Were we not trying to make all our logic to use xhtml that is html5 compatible? Does this not move away from that?

pumazi:
My belief and fear is that we are displacing the issue rather than fixing it. It's like ignoring a foot pain and having it transfer into your back because you inadvertently corrected your posture to deal with the foot pain.

Maybe I just need some reassurance that this isn't going to be a workaround that will sit or displace further. It's not like this issue is blocking content from baking, so I'm not sure the workaround is completely necessary. Please link a github issue stating the complete and root fix to this issue.

cnxepub/collation.py
philschatz: Yeppers, totally agree on the gnarliness of replacing XPath with CSS selectors (I tried doing that the 1st time I noticed XPaths in `cnx-easybake`.

Fortunately, it seems cssselect2 _does_ support namespace prefixes (at least parsing them): https://github.com/Connexions/cssselect2/blob/master/cssselect2/parser.py#L211

reedstrm:
@mulich I'd be happy to add big FIXMEs to the code in these spots. The live CTE blocker bug is:
Connexions/cnx-recipes#179 -  `Multiple Books => Key Terms not sorted alphabetically.` This is due to the use of xpaths in the `sort-by` code in easybake. The right fix is to convert that to use the non-xpath cssselect2 implementation. Which @philschatz and I have added to our list of things to address in the post-CTE easybake rework.

 I threw in the towel on the `sort-by` change after about a day. Phil also says above that he's poked at that once (though without the motivation of fixing a bug - it was just a make the code nice thing) and stopped. I think I see the path to making it work now, but it'd be in the scope of adding namespace awareness to the entire tool (Which Phil seems to be eager to start on :-/ )

 This work around only took a hour to code and a couple more to clean up the tests.  If we don't use it, I want to pull out the change to license metadata anyway - the business logic is wrong.

philschatz:
I believe https://github.com/Connexions/cnx-easybake/pull/72 may fix the issue described here (2hrs of coding + 45min of _manual_ pep8-compliance linting...)

reedstrm:
Closing - we have a potential solution via implementing namespace support in easybake

PR #122, opened 2017-07-11
metadata logic - license and language

reedstrm:
remove some unnecessary namespaces in xpaths used for metadata extraction.  Modified logic for metadata parsing for license* and language - use the `ancestor-or-self` axis in order to inherit certain metadata from parent nodes.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -539,6 +539,7 @@ HTML_DOCUMENT = """\
       xmlns:mod="http://cnx.rice.edu/#moduleIds"
       xmlns:md="http://cnx.rice.edu/mdml"
       xmlns:c="http://cnx.rice.edu/cnxml"
+      lang="{{ metadata['language'] }}"
       >
   <head itemscope="itemscope"
         itemtype="http://schema.org/Book"
--- a/cnxepub/html_parsers.py
+++ b/cnxepub/html_parsers.py
@@ -65,7 +65,7 @@ def _nav_to_tree(root):
             shortid = li.get('cnx-archive-shortid')
             yield {'id': itemid,
                    # Title is wrapped in a span, div or some other element...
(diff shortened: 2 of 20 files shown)


cnxepub/html_parsers.py
```diff
@@ -197,18 +201,34 @@ def keywords(self):
...
     @property
     def license_url(self):
-        items = self.parse('.//xhtml:*[@data-type="license"]/@href')
+        # Three cases for location of metadata stanza:
+        #  1. direct child of current node
+        #  2. direct child of any ancestor
+        #  3. Top of book (occurs when fetching from root)
+        items = self.parse('ancestor-or-self::*/*[@data-type="metadata"]//*'
+                           '[@data-type="license"]/@href'
+                           ' | /xhtml:html/xhtml:body/*'
```
philschatz: isn't `xhtml:body` one of the `ancestor`'s already caught by the previous XPath?

cnxepub/formatters.py
```diff
@@ -539,6 +539,7 @@ def _replace_exercises(elem):
       xmlns:mod="http://cnx.rice.edu/#moduleIds"
       xmlns:md="http://cnx.rice.edu/mdml"
       xmlns:c="http://cnx.rice.edu/cnxml"
+      lang="{{ metadata['language'] }}"
```
philschatz: could/should this be `xml:lang="..."`? I'm not sure when this becomes an issue but I'm guessing it will at least be a problem in the RNG validator.

cnxepub/html_parsers.py
reedstrm: This is case 3 in the comment: when asking the document as a whole what the license is (holding the html root node), the license used to be picked up from the grandchild node (html>body>metadata), but I removed the 'any child' logic, so that specific case (which is reasonable and perhaps common - it's in a least on test) needs to be in there.

Issue #123, opened 2017-08-16
in epub links need file extensions when exported to a epub/zip


reedstrm:
currently internal epub links that point to another file do not contain the file extension that has been written to the filesystem, and so are broken. Also, make'm relative. (currently `/content/<UUID>@<ver>#target` need to become `<UUID>@<ver>.<ext>#target`

Notes from chat:

If you want to take a crack at it, all export_epub does is generate an EPUB from the db, then call to_file on that. At that point, `Items` in the `Package` inside the `EPUB` are serially transformed to HTML, and written to disk. You'd need to wedge a resolver fixup in there - there's already a set of links on the `Item` model, I think. (edited)

brittweinstein:
CLOSING. If this presents itself as an issue again, we will reopen.

PR #124, opened 2017-08-16
maybe fix the inter-module links to contain .xhtml

philschatz:
Work-in-progress. This might also affect publishing so might not be the best place for this fix

Refs #123


brittweinstein:
CLOSING. If this presents itself as an issue again, we will reopen.

PR #125, opened 2017-08-17
rewrite links to other pages in epub

reedstrm:
One more step towards a valid epub. Hopefully useful for TEA importing as well
:skull_and_crossbones: DO NOT MERGE - seems links are _not_ broken on gateway

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -158,6 +158,7 @@ def _make_package(binder):
                 is_navigation=True, properties=['nav'])
     items.append(item)
     resources = {}
+    page_map = {doc.id: doc.ident_hash for doc in flatten_to_documents(binder)}
     # Roll through the model list again, making each one an item.
     for model in flatten_model(binder):
         for resource in getattr(model, 'resources', []):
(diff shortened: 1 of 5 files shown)


cnxepub/adapters.py
```diff
@@ -158,6 +158,7 @@ def _make_package(binder):
                 is_navigation=True, properties=['nav'])
     items.append(item)
     resources = {}
+    page_map = {doc.id: doc.ident_hash for doc in flatten_to_documents(binder)}
```
philschatz: Nice! Could there be a comment that says what this was for (getting the `@{ver}`)?

cnxepub/adapters.py
```diff
@@ -193,6 +194,15 @@ def _make_package(binder):
                 resource = resources.get(filename)
                 if resource:
                     reference.bind(resource, '../resources/{}')
+                else:
+                    if reference.uri.startswith('/contents'):
```
philschatz: Could the `else` case raise an error rather than silently failing?

Aside: Could this test for a uuid regexp? I would really like to get `/contents/` out of the requirements for the HTML and it looks like the resulting link does not contain `/contents/` either

reedstrm: Hmm, nope. We add '/contents' over in adapters, and take it away here. The ID can actually be almost anything (though I think I may have added code that enforces that the ident_hash part has to be uuid@ver ) Also, smallest possible change here, to get it out and released.

cnxepub/adapters.py
reedstrm: Slightly more general than that, but yes, I'll add a comment.

reedstrm:
Yes, yes coverage and tests. I only pushed this at 11pm, after all. :-) Actually it's a bit annoying that the existing tests didn't break, implying not much coverage of this. I can probably tweak the data set for one of the existing tests and get this path covered.

reedstrm:
:skull_and_crossbones: DO NOT MERGE - may not be needed for TEA/TEG

pumazi:
@philschatz, can you revisit this? Have your comments and requests been addressed?

reedstrm:
This is of historical interest only, at this point. Closing.

PR #126, opened 2017-08-31
Remove extra xmlns:bib in sanitize xml XML_WRAPPER

karenc:
We have parser recover=True which means we have hidden some xml errors when
parsing the xml in _sanitize_xml.  This was meant for the input xml but this
hid the error that we have in our code:

```
cnxepub/models.py:96: in _sanitize_xml
    xml = etree.fromstring(XML_WRAPPER.format(raw_xml), xml_parser)
src/lxml/lxml.etree.pyx:3228: in lxml.etree.fromstring (src/lxml/lxml.etree.c:79609)
src/lxml/parser.pxi:1848: in lxml.etree._parseMemoryDocument (src/lxml/lxml.etree.c:119128)
src/lxml/parser.pxi:1736: in lxml.etree._parseDoc (src/lxml/lxml.etree.c:117808)
src/lxml/parser.pxi:1102: in lxml.etree._BaseParser._parseDoc (src/lxml/lxml.etree.c:112052)
src/lxml/parser.pxi:595: in lxml.etree._ParserContext._handleParseResultDoc (src/lxml/lxml.etree.c:105896)
src/lxml/parser.pxi:706: in lxml.etree._handleParseResult (src/lxml/lxml.etree.c:107604)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

E   XMLSyntaxError: Attribute xmlns:bib redefined, line 13, column 38 (line 13)

src/lxml/parser.pxi:635: XMLSyntaxError
```

### DIFF
--- a/cnxepub/models.py
+++ b/cnxepub/models.py
@@ -52,7 +52,6 @@ ATTRIBUTED_ROLE_KEYS = (
 XML_WRAPPER = """\
 <div
   xmlns="http://www.w3.org/1999/xhtml"
-  xmlns:bib="http://bibtexml.sf.net/"
   xmlns:data="http://www.w3.org/TR/html5/dom.html#custom-data-attribute"
   xmlns:epub="http://www.idpf.org/2007/ops"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
@@ -83,9 +82,9 @@ def utf8(item):
         return item
 
 
-def _sanitize_xml(raw_xml):
+def _sanitize_xml(raw_xml, recover=True):
     """Wraps the XML and sanitizes the namespaces."""
-    xml_parser = etree.XMLParser(ns_clean=True, recover=True)
+    xml_parser = etree.XMLParser(ns_clean=True, recover=recover)
     elms = lxml.html.fragments_fromstring(raw_xml)
     # If the raw_xml starts with untagged content, it will be parsed
     # to a string rather than an Element instance.
--- a/cnxepub/tests/test_models.py
+++ b/cnxepub/tests/test_models.py
@@ -129,6 +129,11 @@ class PrivateUtilitiesTestCase(unittest.TestCase):
         xml = target(content)
         self.assertEqual(xml, 'aaa bbb')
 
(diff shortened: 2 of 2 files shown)


karenc:
This fixes the weird bug we have where the image tag at the end of the page lost the angle brackets `<` `>` and became part of the content:

broken-img-tag


https://archive-tea.cnx.org/contents/b1258494-63dd-4a72-b0f0-a09e659964b7@5.209:2mxXJYjP@5

karenc:
This contains the html that breaks sanitize_xml: [easybake_out.zip](https://github.com/Connexions/cnx-epub/files/1266703/easybake_out.zip)

The command line I used for testing:

```
python -c "import cnxepub.collation; f = open('/tmp/easybake_out.html'); html = f.read(); f.close(); import io; binder = cnxepub.collation.reconstitute(io.BytesIO(html)); print(binder[2][6].content)" | uniq | tail -10
```

reedstrm:
Looks good. Is there any places in the code we know we should have good XML, so should pass a `recovery=False`?

karenc:
Yeah, we need more time and discussion to decide on this kind of changes.

cnxepub/models.py
```diff
@@ -83,9 +82,9 @@ def utf8(item):
         return item
 
 
-def _sanitize_xml(raw_xml):
+def _sanitize_xml(raw_xml, recover=True):
```
philschatz: Is there any reason we should use `recover=True`? It seems that by keeping it `recover=True` we are leaving this little landmine/rabbit-hole for future content... right?

karenc: I think we should check what's going to break, instead of just changing it.  I think it can be raised as a separate issue.  I have added this to Connexions/cnx-archive#518 for error handling.

PR #127, opened 2017-11-15
make composite objects track top version, not parent

reedstrm:
The content of any given Composite Page or Binder can change with changes to any part of the book as a whole, not just direct ancestors. Explicitly track that info.


reedstrm:
grr, I guessed coverage would be a problem. Our test data is very … not representative of our actual use-cases, I'm afraid.

kerwinso:
Trello card: https://trello.com/c/jng8rG94

pumazi:
@reedstrm How would you feel about coding object parent relationships instead of passing in parent information? This would enable object traversal which can be coded as methods on the class itself. This proposal would be refactor for some later time, as it is much more involved than what you have here.

PR #128, opened 2017-12-01
verison -> version, and better logic for id and short-id when computi…

reedstrm:
…ng from parent

Should have caught this in the tests, but the story for version vs no version is somewhat … murky.


reedstrm:
The restructing of the order of version vs uri setting is what increased coverage, BTW.

PR #133, opened 2018-12-14
Content truncation fix

reedstrm:
adapting full-html after baking has been mysteriously truncating content with lxml 4.2.5. Fix by removing the support for non-valid-xml-tree documents.

### DIFF
--- a/cnxepub/models.py
+++ b/cnxepub/models.py
@@ -9,16 +9,15 @@ import io
 import hashlib
 import mimetypes
 try:
-    from collections.abc import MutableSequence, Iterable
+    from collections.abc import MutableSequence
 except ImportError:
-    from collections import MutableSequence, Iterable
+    from collections import MutableSequence
 try:
     from urllib.parse import urlparse
 except ImportError:
     from urlparse import urlparse
 from contextlib import contextmanager
 
-import lxml.html
 from lxml import etree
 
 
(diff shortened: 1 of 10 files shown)


reedstrm:
Ah I should drop the knowledge here, just-in-case: all cases of truncation that I found matched the xpath //figure[span[@data-alt and not(node())]] i.e. a figure containing a span with a data-alt attribute, and no children. In all cases, this was the last element on the page (depending on the page, it was nested inside one or more section or div tags - not sure how to write 'last child in the document' as an xpath)

The cause was that we were processing all pages that didn't have a <body> tag as possibly broken HTML, rather than as well-formed XML. This was to enable using HTML fragments (text with some inline markup) as a document body. We ran the parser is "recovery" mode, which is why these bakings silently deleted content, instead of throwing an error, since that mode supports things like forgetting to close tags, etc. This PR removes that support - all our documents are well formed.

cnxepub/tests/test_models.py
```diff
@@ -103,67 +103,6 @@ def test_document_attribs(self):
         self.assertEqual(document.metadata['version'], '2')
 
 
-class PrivateUtilitiesTestCase(unittest.TestCase):
-
-    def test_sanitize_xml(self):
```
BryanEliDimas: Hm... the function being tested here originally was added to fix another bug that was **also** causing missing content. To simply delete that function and these tests without replacing them with other types of tests (like integration tests that somehow test for no missing content due to html fragments or whatever was causing this older bug) seems **too risky** to me; we could end up with missing content again.

cnxepub/models.py
```diff
@@ -82,36 +81,22 @@ def utf8(item):
...
         bod = tree.xpath('//*[self::body|self::x:body]',
                          namespaces={'x': 'http://www.w3.org/1999/xhtml'})[0]
-        bod.tag = 'div'
+        bod.tag = '{http://www.w3.org/1999/xhtml}div'
         return bod
-    # Reeeally broken HTML, or no body
-    except (etree.Error, AttributeError, IndexError):
-        content = _sanitize_xml(content)
-        return lxml.html.fragment_fromstring(content, 'div')
+    except IndexError:
```
BryanEliDimas: Since this is a generic/base exception, it is not 100% clear to me what could be causing it. So I'd suggest adding a comment explaining what might cause it exactly.

cnxepub/tests/test_formatters.py
```diff
@@ -624,20 +624,23 @@ def setUp(self):
 
         metadata = self.base_metadata.copy()
         contents = io.BytesIO(u"""\
+<div>
```
BryanEliDimas: What would happen without adding this line and the closing tag? I assume they are relevant so I'm guessing that the test might fail because it'd be considered broken html, xhtml, or xml. Maybe? And the function being deleted in this PR that "sanitizes" XML was previously making the tests pass?

cnxepub/models.py
karenc: I think this is meant to catch the `[0]` in:

```
        bod = tree.xpath('//*[self::body|self::x:body]',
￼	                         namespaces={'x': 'http://www.w3.org/1999/xhtml'})[0]
```

karenc: Might be better to do `if bod:` instead of try except `IndexError`?

reedstrm: Right, notice that the diff in that case is not adding the catching of the `IndexError`, it's reducing the number of classes of errors caught (from `etree.Error`, `AttributeError`, `IndexError` to just IndexError) I agree it's not the best way to write this code, I can recast it as Karen suggests.

cnxepub/tests/test_formatters.py
reedstrm: Correct - with this PR, we no longer support (and never really used) sending fragments as document content.

reedstrm:
re: adding a test - most of the last 2 weeks was chasing how to get a reproducible test case. The problem is the only case I've found requires baking an entire book, then checking all the baked page outputs with the XPATH I posted above. Even then, which pages are truncated varies per run! (on a busy system that's doing other things: a quiescent system will often repeat the same pages for a few bakes, then change).  My gut tells me we're tickling some sort of memory-overwrite error in the lxml etree HTML recovery code. The bulk of this fix is removing that call, and associated logic.

reedstrm:
(I see my extended comment was missing some quotes, so it was hard to read the `xpath` and the `<body>` tag refs - fixed)

pumazi:
Looks like this does followup on the issues @philschatz reported in https://github.com/Connexions/cnx-epub/pull/131

cnxepub/tests/test_formatters.py
reedstrm: The bulk of these are tests setting up content to test something other than setting up content: they just took advantage of support for HTML fragments (rather than a valid XHTML tree). The code would wrap them in a div, as part of the fragment processing, so I just did that in the tests directly.

reedstrm:
BTW cnx-archive passes it's tests with this in place, as well.

pumazi:
cnx-publishing, which is the more extensive user of cnx-epub, does not pass tests. 😞

PR #137, opened 2019-01-11
remove the extra wrapper node, that we add back as body

reedstrm:
DocumentContentFormatter adds a `body` node, to what it epected was an HTML fragment, Since we now store a rooted tree, this removes the top-level node from that tree, in favor of the added body tag.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -58,6 +58,15 @@ class DocumentContentFormatter(object):
   <body>{}</body>
 </html>""".format(utf8(self.document.content))
         et = etree.HTML(html)
+        # If body now contains a single child, unwrap it
+        body = et.xpath('/html/body')[0]
+        if len(body) == 1:
+            elem = body[0]
+            for child in elem:
+                body.append(child)
+            if elem.text:
+                body.text = elem.text
+            body.remove(elem)
         return etree.tostring(et, pretty_print=True, encoding='utf-8')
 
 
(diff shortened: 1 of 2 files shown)


pumazi:
Use this X when Y if Z, but not when W has P or U, except when Z is a subset of Q or V. 😉

reedstrm:
Yeah, it's tarting to feel like algebra. I'm also not happy with how it's impacting cnx-publishing tests - basically I introduced an asymmetry, so some of the publish tests that create Document objects directly have some non-obvious behaviors.

reedstrm:
I'd rather revert etree_to_content change, to do the unwrap there, but I'm concerned about where else that might affect.

cnxepub/formatters.py
```diff
@@ -58,6 +58,15 @@ def __bytes__(self):
...
 </html>""".format(utf8(self.document.content))
         et = etree.HTML(html)
+        # If body now contains a single child, unwrap it
+        body = et.xpath('/html/body')[0]
+        if len(body) == 1:
+            elem = body[0]
+            for child in elem:
+                body.append(child)
+            if elem.text:
+                body.text = elem.text
```
pumazi: So if there is text similar to `<body><div>foo<p>bar</p></div></body>` we result with `<body><p>bar</p>foo</body>`, no?

reedstrm: body.text is first, so it'll end up `<body>foo<p>bar</p>` what you wrote would have `p.tail = 'foo'`

reedstrm: Right, I added it because the existing test_document test starts as `<p>コンテンツ...</p>`, which is unwrapped to `コンテンツ...'

reedstrm:
Let's hold off on this, I want to try the deeper unwrap I mentioned above.

reedstrm:
ok I tried the revert, and that's not good, either - it's still asymmetric, just in a different way.

cnxepub/formatters.py
pumazi: That's kinda strange that the tail doesn't need to be mentioned and the text does, but for whatever reason it just shows up.

PR #139, opened 2019-01-15
Baking uses the HTMLFormatter - remove the top div there, instead

reedstrm:
Here's the one-more-hot-fix-fix-fix for the extra div that was breaking all the recipes. Local baking test by Helene seems to be positive.

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -21,7 +21,7 @@ from copy import deepcopy
 import requests
 
 from .models import (
-    model_to_tree,
+    model_to_tree, etree_to_content,
     flatten_to_documents,
     Binder, TranslucentBinder,
     Document, DocumentPointer, CompositeDocument, utf8)
@@ -156,11 +156,7 @@ class HTMLFormatter(object):
             else:
                 _html = self.model._xml
 
-            content = ''.join(utf8([
-                               isinstance(node, (type(''), type(b''))) and
-                               node or etree.tostring(node)
-                               for node in _html.xpath('node()')]))
-            return content
+            return etree_to_content(_html, strip_root_node=True)
 
     @property
     def _template(self):
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -159,14 +159,16 @@ class HTMLFormatter(object):
             return tree_to_html(
                 model_to_tree(self.model), self.extensions).decode('utf-8')
         elif isinstance(self.model, Document):
-            content = self.model.content
             if self.generate_ids:
                 _html = deepcopy(self.model._xml)
                 self._generate_ids(self.model, _html)
-                content = ''.join(utf8([
-                                   isinstance(node, (type(''), type(b''))) and
-                                   node or etree.tostring(node)
-                                   for node in _html.xpath('node()')]))
+            else:
+                _html = self.model._xml
+
+            content = ''.join(utf8([
+                               isinstance(node, (type(''), type(b''))) and
+                               node or etree.tostring(node)
+                               for node in _html.xpath('node()')]))
             return content
 
     @property
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -58,15 +58,6 @@ class DocumentContentFormatter(object):
   <body>{}</body>
 </html>""".format(utf8(self.document.content))
         et = etree.HTML(html)
-        # If body now contains a single child, unwrap it
-        body = et.xpath('/html/body')[0]
-        if len(body) == 1:
-            elem = body[0]
-            for child in elem:
-                body.append(child)
-            if elem.text:
-                body.text = elem.text
-            body.remove(elem)
         return etree.tostring(et, pretty_print=True, encoding='utf-8')
 
 
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -21,7 +21,7 @@ from copy import deepcopy
 import requests
 
 from .models import (
-    model_to_tree, etree_to_content,
+    model_to_tree, content_to_etree, etree_to_content,
     flatten_to_documents,
     Binder, TranslucentBinder,
     Document, DocumentPointer, CompositeDocument, utf8)
@@ -55,9 +55,10 @@ class DocumentContentFormatter(object):
     def __bytes__(self):
         html = """\
 <html xmlns="http://www.w3.org/1999/xhtml">
-  <body>{}</body>
+  {}
 </html>""".format(utf8(self.document.content))
-        et = etree.HTML(html)
+        html = _fix_namespaces(html.encode('utf-8'))
+        et = etree.HTML(html.decode('utf-8'))
         return etree.tostring(et, pretty_print=True, encoding='utf-8')
 
 
@@ -150,11 +151,9 @@ class HTMLFormatter(object):
             return tree_to_html(
                 model_to_tree(self.model), self.extensions).decode('utf-8')
         elif isinstance(self.model, Document):
+            _html = content_to_etree(self.model.content)
             if self.generate_ids:
-                _html = deepcopy(self.model._xml)
                 self._generate_ids(self.model, _html)
-            else:
-                _html = self.model._xml
 
             return etree_to_content(_html, strip_root_node=True)
 
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,10 +5,10 @@
 
    - feature message
 
-0.16.2
+0.17.0
 ------
 
-- Replicate earlier behavior by removing top node when serializing to HTML
+- Limit Document content to <body> tags only
 
 0.16.1
 ------
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -524,7 +524,9 @@ def _adapt_single_html_tree(parent, elem, nav_tree, top_metadata,
                 if key in ('itemtype', 'itemscope'):
                     child.attrib.pop(key)
 
-            contents = etree.tostring(child)
+            document_body = content_to_etree('')
+            document_body.append(child)
+            contents = etree.tostring(document_body)
             model = {
                 'page': Document,
                 'composite-page': CompositeDocument,
--- a/cnxepub/models.py
+++ b/cnxepub/models.py
@@ -97,7 +97,12 @@ def content_to_etree(content):
         return tree
 
 
-def etree_to_content(etree_):
+def etree_to_content(etree_, strip_root_node=False):
+    if strip_root_node:
+        return ''.join(utf8([
+            isinstance(node, (type(''), type(b''))) and
+            node or etree.tostring(node)
+            for node in etree_.xpath('node()')]))
     return etree.tostring(etree_)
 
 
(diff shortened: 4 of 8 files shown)


reedstrm:
Further testing showed that the formatting steps in the baking pipeline gt run one-after-the-other, so removing the extra top-div is definitely needed at the HTMLFormatter step, it is now _not_ needed at the DocumentFormatter step: the content after baking is juuust right. So, putting _back_ the previous code for that step.

cnxepub/formatters.py
```diff
@@ -159,14 +159,16 @@ def _content(self):
...
                 _html = deepcopy(self.model._xml)
                 self._generate_ids(self.model, _html)
-                content = ''.join(utf8([
-                                   isinstance(node, (type(''), type(b''))) and
-                                   node or etree.tostring(node)
-                                   for node in _html.xpath('node()')]))
+            else:
+                _html = self.model._xml
+
+            content = ''.join(utf8([
```
karenc: At first glance this kind of makes sense, since `self._generate_ids()` only changes ids and doesn't add or remove tags so whether we generate ids or not, we should have the same code to generate the content.

Then I went back to look at why we have code to strip the root tag, and it's because we [used to create an extra div element](https://github.com/openstax/cnx-epub/commit/bf5fbc41bf8f4a68a73099b06b5f6b6e697a6060#diff-6599de93a091ed0fb6b2c7b619b19c48R142):

```python
            if self.generate_ids:
                div = lxml.html.fragment_fromstring(content, 'div')
                self._generate_ids(self.model, div)
                content = ''.join(utf8([
                    isinstance(node, (type(''), type(b''))) and
                    node or etree.tostring(node)
                    for node in div.xpath('node()')]))
```

At some point [we changed it](https://github.com/openstax/cnx-epub/commit/333141fe48110a123c2f0a84e5d4d647813ab19f#diff-6599de93a091ed0fb6b2c7b619b19c48L144) to not add a div but still remove the root tag:

```diff
@@ -141,12 +143,12 @@ class HTMLFormatter(object):
         elif isinstance(self.model, Document):
             content = self.model.content
             if self.generate_ids:
-                div = lxml.html.fragment_fromstring(content, 'div')
-                self._generate_ids(self.model, div)
+                _html = deepcopy(self.model._xml)
+                self._generate_ids(self.model, _html)
                 content = ''.join(utf8([
-                    isinstance(node, (type(''), type(b''))) and
-                    node or etree.tostring(node)
-                    for node in div.xpath('node()')]))
+                                   isinstance(node, (type(''), type(b''))) and
+                                   node or etree.tostring(node)
+                                   for node in _html.xpath('node()')]))
             return content
 
     @property
```

At this point, we have different output depending on whether `self.generate_ids` is true.  If `self.generate_ids` is true, the root tag is removed, if false, the content is unchanged.

We need to have a look at [self.model._xml](https://github.com/openstax/cnx-epub/blob/326e013975183ef97bb6bea84607aafc510bca97/cnxepub/models.py#L85) to see if we should always remove the root tag:

```python
def content_to_etree(content):
    if not content:  # Allow building empty models
        return etree.XML('<div xmlns="http://www.w3.org/1999/xhtml" />')
    xml_parser = etree.XMLParser(ns_clean=True)
    tree = etree.XML(content, xml_parser)
    # Determine if we've been fed a full XHTML page, with a <body> tag:
    bods = tree.xpath('//*[self::body|self::x:body]',
                      namespaces={'x': 'http://www.w3.org/1999/xhtml'})
    if bods:
        bods[0].tag = '{http://www.w3.org/1999/xhtml}div'
        return bods[0]
    else:  # It parsed, so must have been fed a valid XML tree - return it
        return tree
```

If content was empty, we add a `<div>` tag, if the content has a body tag, we change the body tag to a div, otherwise no change.  So I'm not sure if it's correct to just remove the root tag.

---

The other thing I'm concerned about is... we need to be able to create a `Binder`, use `SingleHTMLFormatter()` to convert it into a single html, and then get the `Binder` back by calling `cnxepub.adapters.adapt_single_html`.  I wrote some code to do this and this is the result:

```
content: <p>hello</p>
document content in binder: <p>hello</p>
adapted content: <div xmlns="http://www.w3.org/1999/xhtml" data-type="page" id="ingress"/>

content: <html><body>hello</body></html>
document content in binder: <html:div xmlns:html="http://www.w3.org/1999/xhtml">hello</html:div>
adapted content: <div xmlns="http://www.w3.org/1999/xhtml" data-type="page" id="ingress"/>

content: <html><body><p>hello</p></body></html>
document content in binder: <html:div xmlns:html="http://www.w3.org/1999/xhtml"><p>hello</p></html:div>
adapted content: <div xmlns="http://www.w3.org/1999/xhtml" data-type="page" id="ingress"><p id="39337">hello</p>
  </div>
```

In the first and second cases, the content is gone.. The third case is may be correct?

This is the code I wrote:

```python
# -*- coding: utf-8 -*-       
                              
import io

from cnxepub.adapters import adapt_single_html
from cnxepub.formatters import SingleHTMLFormatter
from cnxepub.models import Binder, Document


def format_and_adapt(content):
    # build a binder with one page
    base_metadata = {
        'publishers': [],
        'created': '2013/03/19 15:01:16 -0500',
        'revised': '2013/06/18 15:22:55 -0500',
        'authors': [
            {'type': 'cnx-id',
             'name': 'Sponge Bob',
             'id': 'sbob'}],
        'editors': [],
        'copyright_holders': [],
        'illustrators': [],
        'subjects': ['Science and Mathematics'],
        'translators': [],
        'keywords': ['Bob', 'Sponge', 'Rock'],
        'title': "Goofy Goober Rock",
        'license_text': 'CC-By 4.0',
        'license_url': 'http://creativecommons.org/licenses/by/4.0/',
        'summary': "<p>summary</p>",
        'version': 'draft',
    }

    metadata = base_metadata.copy()
    metadata.update({'title': 'Kraken'})
    binder = Binder('Kraken', metadata=metadata)
    metadata = base_metadata.copy()
    metadata.update({'title': 'entrée'})
    binder.append(Document('ingress', io.BytesIO(content),
                  metadata=metadata))

    single_html = str(SingleHTMLFormatter(binder))
    adapted_binder = adapt_single_html(single_html)
    print('content: {}'.format(content))
    print('document content in binder: {}'.format(binder[0].content))
    print('adapted content: {}\n'.format(adapted_binder[0].content))


format_and_adapt(b'<p>hello</p>')
format_and_adapt(b'<html><body>hello</body></html>')
format_and_adapt(b'<html><body><p>hello</p></body></html>')
```

reedstrm: Good point, I agree on a general "this is very surprising, what the heck is the model doing" way. Stripping out fragment support like I did was a bit … extreme, it seems.
On the other hand, I'm not convinced any actual production workflow is putting anything like case 1 or 2 into a Document - all the pages are full HTML pages initially created from CNXML, with the `<body>` tag converted from the `<content>` tag of the CNXML. Free text is not allowed as a child of `<content>`, so all text will always be wrapped in at least a `<p>`. I think I can write a quick `xpath` test in the DB to confirm that.

reedstrm: The thornier case may be what is baking creating, since that output is adapted from a single html to create pages. Yesterday's code was in fact stripping an extra div from the top - when going to the database (via DocumentFormater)

karenc: What we can do is to add a `<div>` to `<p>hello</p>` in `content_to_etree()`, that will make everything correct then? :thinking:

reedstrm: (Love this test code BTW) I just rolled back to pre-fragment-support removal, and case 2 still deletes the content. Case 1 now passes through. Hmm.

reedstrm: I added a DocumentContentFormatter output to your test. It's now even more confusing:
```$ python test.py
content: <p>hello</p>
document content in binder: <p>hello</p>
adapted content: <div xmlns="http://www.w3.org/1999/xhtml" data-type="page" id="ingress"/>
formatted content: <html xmlns="http://www.w3.org/1999/xhtml">
  <body><p>hello</p></body>
</html>


content: <html><body>hello</body></html>
document content in binder: <html:div xmlns:html="http://www.w3.org/1999/xhtml">hello</html:div>
adapted content: <div xmlns="http://www.w3.org/1999/xhtml" data-type="page" id="ingress"/>
formatted content: <html xmlns="http://www.w3.org/1999/xhtml">
  <body><div xmlns:html="http://www.w3.org/1999/xhtml">hello</div></body>
</html>


content: <html><body><p>hello</p></body></html>
document content in binder: <html:div xmlns:html="http://www.w3.org/1999/xhtml"><p>hello</p></html:div>
adapted content: <div xmlns="http://www.w3.org/1999/xhtml" data-type="page" id="ingress"><p id="61569">hello</p>
  </div>
formatted content: <html xmlns="http://www.w3.org/1999/xhtml">
  <body><div xmlns:html="http://www.w3.org/1999/xhtml"><p>hello</p></div></body>
</html>
```

reedstrm: The missing content has reappeared, though cases 2 and 3 now have an extra top-level div.

reedstrm: I thought about adding a `div` unilaterally, as well - the old code basically parsed the input via HTML parser, in recovery mode, which adds <html><body> to the doc, if it doesn't already have that, then found the contents of the body. The original missing content bug, though, seemed to be coming from that HTML parser (and recovery mode, in particular)

karenc: This test code should be part of the `adapt_single_html` test suite.  I didn't think to add it when I originally wrote the code. :disappointed:

reedstrm: 100% agreed

karenc:
@reedstrm I added 2 commits to your branch.  It passes the cnx-epub tests but I haven't checked cnx-publishing and cnx-archive yet so going to do that now.

reedstrm:
ok. I think I see what you did here. I'm running your test script (w/ my addition of a case w/ content = `<para><p>hello</p></para>` to emulate the after-baking case (not using `div` because I want to see what's added vs. passed through), and I'm concerned that we'll see the return of the extra wrapper-div in the baked path.

reedstrm:
I think what I want to do is write some tests that step through the same content transforms as a book-bake: create a binder like it would be build from the DB, then transform (collate) w/ a recipe, and reparse out as if saving to DB. We're missing that full-path test, I guess.

karenc:
> I think what I want to do is write some tests that step through the same content transforms as a book-bake: create a binder like it would be build from the DB, then transform (collate) w/ a recipe, and reparse out as if saving to DB. We're missing that full-path test, I guess.

I think we should have that in cnx-publishing

karenc:
I've decided to remove support for non-body root tag for `Document`.  Hope that makes things less complicated... :crossed_fingers:

reedstrm:
> I've decided to remove support for non-body root tag for `Document`. Hope that makes things less complicated...

Umm, post-baking we do exactly that, for every page. _ goes and looks at the code_ Ah, you override the tag to _be_ a body tag in those cases, I see. Hmm.

karenc:
@reedstrm I changed the code a bit in adapters.py because we need `<div data-type="page">` in the body when we create the documents.

@helenemccarron confirmed that https://devb.cnx.org/contents/1Q9uMg_a@10.1:8yFBtNIX@10.1/Chapter-Review looks correct so it's probably working :crossed_fingers: 

Could you have a look at the changes and see if they're ok to go out?

PR #146, opened 2019-03-20
Fix and make the XML squashing function reusable

pumazi:
## What Where Why?

This function plays a huge role in our system because it is responsible for manipulating our content's title and summary (aka abstract). This tests it at a unit-testing level, probably for the first time ever. I do believe it may actually be catching possible future bugs.

## Description

I'm putting this function in a reusable place so that it can be used within Nebuchadnezzar. I've taken liberties to fix many problems with it along the way. From duplicating ending text, so not html entity encoding text that doesn't have html.

### UTF-8 instead of XML entities

The decision to encode all text as UTF-8 was made. This means XML entities no longer appear in the resulting snippet. This was happening in several places within the tests. At the same time it wasn't happening in other places. Webview does have the `<meta charset="utf-8">` in the page, so we should be fine to support straight up UTF-8 text. Worst case, we can switch it back (see my second commit).

It does bring up a point that I'm sure Phil has made several times over. That is, we don't have a unified way of parsing our XML. That should really be addressed.

### Leading and trailing whitespace

This removes white space around the content. This wasn't necessarily a problem, but it does now provide consistent output.

### Now handles math

I didn't see any tests that specifically looked at math in relation to this squashing functionality. Those tests are now included and seem to be handled correctly with the absence of a parent element.

### DIFF
--- a/cnxepub/html_parsers.py
+++ b/cnxepub/html_parsers.py
@@ -10,23 +10,13 @@ import re
 from lxml import etree
 
 from .models import TRANSLUCENT_BINDER_ID
+from cnxepub.utils import squash_xml_to_text
 
 
 HTML_DOCUMENT_NAMESPACES = {
     'xhtml': "http://www.w3.org/1999/xhtml",
     'epub': "http://www.idpf.org/2007/ops",
     }
-
-
-def _squash_to_text(elm, remove_namespaces=False):
-    value = [elm.text or '']
-    for child in elm.getchildren():
-        value.append(etree.tostring(child).decode('utf-8').strip())
-        value.append(child.tail or '')
-    if remove_namespaces:
-        value = [re.sub(' xmlns:?[^=]*="[^"]*"', '', v) for v in value]
-    value = ''.join(value)
-    return value
 
 
 def parse_navigation_html_to_tree(html, id):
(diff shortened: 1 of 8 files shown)


cnxepub/html_parsers.py
```diff
@@ -18,17 +19,6 @@
     }
 
 
-def _squash_to_text(elm, remove_namespaces=False):
-    value = [elm.text or '']
-    for child in elm.getchildren():
-        value.append(etree.tostring(child).decode('utf-8').strip())
-        value.append(child.tail or '')
```
pumazi: Notice this was removed in the new version. In the unit-tests it was duplicating the trailing text.

pumazi:
Based on feedback from Helene, I'll be adjusting this to not encode UTF-8 as XML entities.

cnxepub/tests/test_utils.py
```diff
@@ -0,0 +1,116 @@
...
+    def test_with_namespace_removal(self):
+        result = self.target(self.content, True)
+
+        expected = 'FOO <p><math/></p><p><math/></p> BAR'
+        assert result == expected
+
+    def test_text_only(self):
+        txt = '<div>Hello Wórld!</div>'
+        expected = 'Hello Wórld!'
+        if IS_PY2:
```
karenc: Next time you can just do `txt = u'<div>Hello Wórld!</div>'` so the meaning is the same in both python 2 and 3 instead of doing `.decode('utf-8')` if it's python 2.  Alternatively we can also do `from __future__ import unicode_literals`.

PR #149, opened 2019-07-23
Change _fix_namespaces to not remove xml comments

karenc:
The xml comments, `<!-- no-selfclose -->` are added into
rhaptos.cnxmlutils in order to avoid self closing tags in our html, in
an attempt to make the xhtml output somewhat HTML5 compatible.

`cnxepub.formatters._fix_namespaces` tidies up the namespaces by putting
everything in the root tag.  It uses the python built-in
`xml.etree.ElementTree` library to do that.  The `xml.etree.ElementTree`
library skips xml comments.

From the python docs
https://docs.python.org/3.7/library/xml.etree.elementtree.html#module-xml.etree.ElementTree
(the 2.7 docs doesn't have this):

> Note: Not all elements of the XML input will end up as elements of the
> parsed tree. Currently, this module skips over any XML comments,
> processing instructions, and document type declarations in the input.

So switch from `xml.etree.ElementTree` to `lxml.etree` to preserve xml
comments.

`lxml.etree` also has `register_namespace`, but it doesn't allow a
default namespace.  So what I've done is to parse the document, create a
new root tag that's exactly the same as the old one but with the right
nsmap.

https://stackoverflow.com/questions/11346480/lxml-add-namespace-to-input-file

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -11,6 +11,7 @@ import json
 import logging
 import random
 import sys
+from io import BytesIO
 
 import re
 import jinja2
@@ -306,7 +307,8 @@ class SingleHTMLFormatter(object):
 
 
 def _fix_namespaces(html):
-    nsmap = {u"": u"http://www.w3.org/1999/xhtml",
+    # Get rid of unused namespaces and put them all in the root tag
+    nsmap = {None: u"http://www.w3.org/1999/xhtml",
              u"m": u"http://www.w3.org/1998/Math/MathML",
              u"epub": u"http://www.idpf.org/2007/ops",
              u"rdf": u"http://www.w3.org/1999/02/22-rdf-syntax-ns#",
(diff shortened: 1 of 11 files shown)


cnxepub/formatters.py
```diff
@@ -321,16 +321,17 @@ def _fix_namespaces(html):
...
-    et.write(new_html)
-    new_html.seek(0)
-    let = etree.parse(new_html)
-    return etree.tostring(let, pretty_print=True, encoding='utf-8')
+    # We can't use lxml.etree.register_namespace because it
+    # doesn't allow a default namespace and we can't use
+    # xml.etree.ElementTree.register_namespace because it
+    # removes xml comments which the self closing tag fix is
+    # dependent on.
+    # https://stackoverflow.com/questions/11346480/
```
therealmarv: great for future reference!

cnxepub/tests/data/desserts-includes-py2.xhtml
```diff
@@ -1,6 +1,18 @@
...
+  xmlns:data='http://www.w3.org/TR/html5/dom.html#custom-data-attribute'
+  xmlns:datadev='http://dev.w3.org/html5/spec/#custom'
+  xmlns:dc='http://purl.org/dc/elements/1.1/'
+  xmlns:epub='http://www.idpf.org/2007/ops'
+  xmlns:lrmi='http://lrmi.net/the-specification'
+  xmlns:m='http://www.w3.org/1998/Math/MathML'
+  xmlns:md='http://cnx.rice.edu/mdml'
+  xmlns:mod='http://cnx.rice.edu/#moduleIds'
+  xmlns:qml='http://cnx.rice.edu/qml/1.0'
+  xmlns:rdf='http://www.w3.org/1999/02/22-rdf-syntax-ns#'
```
pumazi: This is a regression, no?

karenc: I don't think that was exactly what we were doing with `_fix_namespace`.  We were trying to put all the namespace in the root tag.  I think I can use `register_namespace` and this together ... :thinking:

karenc: Ok, I think I've fixed it now!

PR #152, opened 2019-10-09
add stimulus to webview exercise template

brittany-johnson:
https://github.com/openstax/webview/issues/2274

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -459,6 +459,7 @@ def exercise_callback_factory(match, url_template,
 #  src/scripts/modules/media/embeddables/exercise-template.html
 
 EXERCISE_TEMPLATE = jinja2.Template("""\
+{{data['items'].stimulus_html}}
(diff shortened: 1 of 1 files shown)


cnxepub/formatters.py
```diff
@@ -459,6 +459,7 @@ def _replace_exercises(elem):
 #  src/scripts/modules/media/embeddables/exercise-template.html
 
 EXERCISE_TEMPLATE = jinja2.Template("""\
+{{data['items'].stimulus_html}}
```
karenc: This needs to include `.0` to indicate the first item.  I tested this by doing:

```
>>> from cnxepub.formatters import EXERCISE_TEMPLATE
>>> EXERCISE_TEMPLATE.render(data={'items': [{'stimulus_html': '<a href="/">test</a>'}]})
'\n'
```

We need to get the first item of `data['items']`.  If we change it to `{{ data['items'].0.stimulus_html }}`, it should work :+1:

PR #153, opened 2019-10-09
Add exercise stimulus if provided

tomjw64:
This change adds to the template for exercises pulled from a request in order to display an exercise stimulus for a multi-part exercise if one is provided.

I also added an exercise stimulus to one of the two existing tests that pull the json.


tomjw64:
@karenc I saw travis failed and went back to change. Let's try again!

PR #159, opened 2020-01-13
Fix composite-page ids erroring in epub

tomjw64:
This should fix the errors that occur when adding ids to composite materials
during baking, meaning we can again try to link to those materials in
our PDF tables of contents. This is acheived by using the attribute
`data-uuid-key` preferentially over the `id` attribute for creating the
uuid for the new module. Now, if composite material has a
`data-uuid-key`, then it should be able to have `id` be added to it safely.

I verified this work by 1) modifying a test, adding an id to a
composite-page and ensuring it failed the test before the change in this
PR and passed after the change in this PR and 2) spinning up publishing
locally and ensuring that statistics successfully baked using a
statistics recipe built with `give-composites-ids` enabled (which was
the type of action which originally alerted us of this issue).

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -494,7 +494,11 @@ def _adapt_single_html_tree(parent, elem, nav_tree, top_metadata,
                 elif parent.metadata.get('version') is not None:
                     metadata['version'] = parent.metadata['version']
 
-            id_ = metadata.get('cnx-archive-uri') or child.attrib.get('id')
+            uuid_key = child.get('data-uuid-key')
+            child_id = child.attrib.get('id')
+            id_ = metadata.get('cnx-archive-uri') or (child_id
+                                                      if not uuid_key
+                                                      else None)
             if not id_:
                 id_ = _compute_id(parent, child, metadata.get('title'))
                 if metadata.get('version'):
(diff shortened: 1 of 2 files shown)


cnxepub/adapters.py
```diff
@@ -494,7 +494,11 @@ def _compute_shortid(ident_hash):
                 elif parent.metadata.get('version') is not None:
                     metadata['version'] = parent.metadata['version']
 
-            id_ = metadata.get('cnx-archive-uri') or child.attrib.get('id')
+            uuid_key = child.get('data-uuid-key')
```
philschatz: Just confirming that these changes only care about the presence of `data-uuid-key`, not the value of `data-uuid-key`.

It seems that the value is used a few lines above at line 442.

tomjw64: Kinda. Although the reason I checked falseyness rather than noneness was because accepting `data-uuid-key` as empty string is probably also a formula for unhappiness.

tomjw64: But if you mean present as in 'a string with positive length', then yes.

philschatz: I was just surprised that the value of `uuid_key` was pulled out of the element but not being used (it seems to be because the value is pulled out of the element in earlier code)

tomjw64: ah, it is being used on line 500 to check for falseyness of the value

PR #160, opened 2020-05-01
Tentative resolution for cnx#924

tomjw64:
I don't really have an idea of how dangerous this change is. Even though it passes all cnx-epub unit tests, I think this probably needs more thorough regression testing, since it potentially changes how all content is serialized. But it resolves https://github.com/openstax/cnx/issues/924 without libxml doing error recovery, which is nice.


tomjw64:
Closing in favor o #161

PR #161, opened 2020-05-05
Update pins due to high priority failures

tomjw64:


setup.py
```diff
@@ -9,7 +9,13 @@
...
 install_requires = [
     'jinja2',
-    'lxml',
+    'lxml==4.4.3',
+    'requests',
+    'python-memcached'
+    ] if IS_PY3 else [
+    'jinja2==2.11.2',
+    'MarkupSafe==1.1.1',
+    'lxml==4.4.3',
```
philschatz, line 18: Can you add a comment in here for why this is pinned to 4.4.3?

c0ldcalamari: This change/pinned lxml to 4.4.3 fixes the following issue with our Polish Books as well https://app.zenhub.com/workspaces/content-engineering-tech-team-5af1f4cc12da5e6d74331b60/issues/openstax/cnx/1165 

Where same fix / pinned version was given to easy-bake https://github.com/openstax/cnx-easybake/pull/115/files for resolution.

PR #163, opened 2020-05-27
adds aside to the elements that get updated ids

philschatz:
This adds `aside` to the whitelist but I am not sure why there was a whitelist to begin with (and a randomly-generated id in the nearby code). #70 added the original code and referred to easybake also generating `auto_{}_{}` ids. 

Maybe since we have a way to build the books easily (bakery CLI) it would be a good time to see if removing the whitelist & the random-id code was ever useful.

The refs openstax/cnx-recipes#2336

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -105,7 +105,7 @@ class HTMLFormatter(object):
         elements = [
             'p', 'dl', 'dt', 'dd', 'table', 'div', 'section', 'figure',
             'blockquote', 'q', 'code', 'pre', 'object', 'img', 'audio',
-            'video',
+            'video', 'aside'
             ]
         elements_xpath = '|'.join(['.//{}|.//xhtml:{}'.format(elem, elem)
                                   for elem in elements])


rnathuji:
@philschatz - Just to check my own understanding, I think in our pipelines this auto generated ID bit gets invoked during [`neb assemble`](https://github.com/openstax/nebuchadnezzar/blob/ca6651c79a72b95982974f86da812c288152139b/nebu/cli/assemble.py#L24) by virtue of [this loop](https://github.com/openstax/cnx-epub/blob/6b6b522014742d92600ac29f3ba106bbb14fc7cc/cnxepub/formatters.py#L260). So is the question we want to explore what the impact would be if we set that to `False` (or otherwise disable the random ID generation) and instead try an alternative mechanism for unique IDs?

philschatz:
Not quite. I would just like to delete these lines https://github.com/openstax/cnx-epub/blob/2cfb6ab3f21b4f8d84692d5051a71e6a56b6b475/cnxepub/formatters.py#L130-L135

Then we would never generate random ids. I think that the only reasons we generate random ids is in case the original document did not have ids on the elements. In order for links in the original content to work the target would already need to have an id. 

Therefore this function should only generate random ids for elements that are not link targets.

In that case we probably don't need id attributes but if we did then we would not want them to change whenever we bake the content.

rnathuji:
Would a useful exercise then be to take the Micro book, remove the random ID insertion, generate the XHTML, and see if it passes the [validation tool you wrote](https://github.com/philschatz/xhtml-validator)? If that looks good we can then see about removing the whitelist altogether?

Also, is this tied to or independent of the PR? (e.g. is approval of this quick change blocking things, or did you want to see how these other options pan out to decide if an alternate implementation is preferable?)

PR #164, opened 2020-05-27
generate deterministic uuids for composite pages

philschatz:
It was pointed out that in some cases a composite page gets a different UUID every time it executes.

This may fix the problem.

### DIFF
--- a/cnxepub/adapters.py
+++ b/cnxepub/adapters.py
@@ -351,7 +351,10 @@ def adapt_single_html(html):
     html_root = etree.fromstring(html)
 
     metadata = parse_metadata(html_root.xpath('//*[@data-type="metadata"]')[0])
-    id_ = metadata['cnx-archive-uri'] or 'book'
+    id_ = metadata['cnx-archive-uri']
+    if id_ is None:
+        raise ValueError('Metadata element did not contain a cnx-archive-uri', 
+            etree.tostring(html_root.xpath('//*[@data-type="metadata"]')[0]))
 
     binder = Binder(id_, metadata=metadata)
     nav_tree = parse_navigation_html_to_tree(html_root, id_)
@@ -437,7 +440,7 @@ def _adapt_single_html_tree(parent, elem, nav_tree, top_metadata,
             except ValueError:
                 pass
         else:  # Punt - no parent uuid, make one up for child
-            return str(uuid.uuid4())
+            raise ValueError('Could not find parent UUID while trying to generate a UUID', p_ids)
 
         uuid_key = elem.get('data-uuid-key', elem.get('class', key))
         if (sys.version_info.major == 2):  # https://bugs.python.org/issue34145
--- a/cnxepub/tests/data/desserts-single-page-bad.xhtml
+++ b/cnxepub/tests/data/desserts-single-page-bad.xhtml
@@ -18,7 +18,7 @@
   <body xmlns:bib="http://bibtexml.sf.net/" xmlns:data="http://www.w3.org/TR/html5/dom.html#custom-data-attribute" itemscope="itemscope" itemtype="http://schema.org/Book">
     <div data-type="metadata">
       <h1 data-type="document-title" itemprop="name">Desserts</h1>
-
+      <span data-type="cnx-archive-uri" data-value="00000000-0000-0000-0000-000000000000"></span>
       <div class="authors">
         By:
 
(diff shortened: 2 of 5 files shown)


cnxepub/adapters.py
```diff
@@ -351,7 +351,10 @@ def adapt_single_html(html):
     html_root = etree.fromstring(html)
 
     metadata = parse_metadata(html_root.xpath('//*[@data-type="metadata"]')[0])
-    id_ = metadata['cnx-archive-uri'] or 'book'
+    id_ = metadata['cnx-archive-uri']
+    if id_ is None:
+        raise ValueError('Metadata element did not contain a cnx-archive-uri',
```
tomjw64, line 356: Could we clarify the failure case here? Is this when a collection does not have a `cnx-archive-uri` metadata entry? How common is this?

cnxepub/adapters.py
```diff
@@ -437,7 +440,7 @@ def _compute_id(p, elem, key):
             except ValueError:
                 pass
         else:  # Punt - no parent uuid, make one up for child
-            return str(uuid.uuid4())
+            raise ValueError('Could not find parent UUID while trying to generate a UUID', p_ids)
```
tomjw64, line 443: My hunch is that this is safe if we can sort out the above, so we can avoid regression with top level pages.

cnxepub/adapters.py
philschatz: It occurred in the tests but I am not sure if it actually occurs in books. I can remove this if this does not arise when baking any of the books.

rnathuji:
I built a distribution pipeline using this branch for neb and cops-bakery-scripts (specifically images `rnathuji/nebuchadnezzar:cnxepub-4b570b6` and `rnathuji/cops-bakery-scripts:cnxepub-4b570b6`). I ran a job with `{"collection_id": "col12087", "server": "staging.cnx.org", "style": "microbiology", "version": "latest"}` and got the following error during `assemble book metadata` where it tries to invoke `reconstitute` on the assembled XHTML:

```
+ exec
++ tee assembled-book-metadata/stderr
++ cat book/collection_id
+ collection_id=col12087
+ book_dir=assembled-book/col12087
+ target_dir=assembled-book-metadata/col12087
+ mkdir assembled-book-metadata/col12087
+ python /code/scripts/assemble-book-metadata.py assembled-book/col12087 assembled-book-metadata/col12087/collection.assembled-metadata.json
Traceback (most recent call last):
  File "/code/scripts/assemble-book-metadata.py", line 48, in <module>
    main()
  File "/code/scripts/assemble-book-metadata.py", line 28, in main
    binder = reconstitute(in_file)
  File "/usr/local/lib/python3.7/site-packages/cnxepub/collation.py", line 51, in reconstitute
    return adapt_single_html(xhtml)
  File "/usr/local/lib/python3.7/site-packages/cnxepub/adapters.py", line 363, in adapt_single_html
    _adapt_single_html_tree(binder, body[0], nav_tree, top_metadata=metadata)
  File "/usr/local/lib/python3.7/site-packages/cnxepub/adapters.py", line 506, in _adapt_single_html_tree
    id_ = _compute_id(parent, child, metadata.get('title'))
  File "/usr/local/lib/python3.7/site-packages/cnxepub/adapters.py", line 443, in _compute_id
    raise ValueError('Could not find parent UUID while trying to generate a UUID', p_ids)
ValueError: ('Could not find parent UUID while trying to generate a UUID', ['col12087', 'col12087'])
```

This is reproducible outside of a pipeline:
```
neb get -r -m -d col12087/raw staging.cnx.org col12087 latest
neb assemble col12087/raw col12087
```
...and then within an appropriate venv:
```
from cnxepub.collation import reconstitute
reconstitute(open("col12087/collection.assembled.xhtml"))
```

tomjw64:
@rnathuji Could you find the offending module/page?

rnathuji:
Seems to be the Introduction (`m58780`) (that's the `elem` input when it fails). I've got it pulled up in the debugger and can show you when we chat in a bit as well.

rnathuji:
@tomjw64 and I looked at the failure case with `col12087`, and it seems the composite pages may be a red herring in that:

* The input to the [uuid()](https://github.com/openstax/cnx-epub/blob/6b6b522014742d92600ac29f3ba106bbb14fc7cc/cnxepub/adapters.py#L435) call ends up being a collection ID string (e.g. `col12087`) instead of an expected UUID format. This cascades into changing UUIDs for composite pages, but solving the problem for raw XHTML should (we believe) solve this for baked XHTML cases
* It seems we need the UUID data for the book in the content, which is not available currently. The unknown is how the value should get into the content.

tomjw64:
Oh, and in case it is desired for checking anything out, `e42bd376-624b-4c0f-972f-e0c57998e765` is the UUID I believe we would want for this collection.

philschatz:
[Issue](https://github.com/openstax/cnx/issues/1011)

Instead, these PR's address the issue: https://github.com/openstax/nebuchadnezzar/pull/182 and https://github.com/openstax/output-producer-service/pull/115

PR #165, opened 2020-05-29
Ensure all IDs are unique and remove randomly-generated IDs

philschatz:
This is a followup to #163 . It removes the whitelist and the non-deterministic random id generation code.


tomjw64:
Now that book builds are more deterministic, could we show that removing id-generation leads to no change in the structure(other than the id attributes, of course) of each book in our library.

brittweinstein:
@philschatz please create a card for this and attach this PR.

m1yag1 (approved):

sparksam (approved):

PR #167, opened 2020-07-17
Apply auto id prefixes to all elements with ids

tomjw64:
Hopefully this should take care of our duplicate id issues for good, at least with respect to cnx-epub.
### In This PR:
- Still only applies **new** IDs to elements in the hard coded list, but now also applies prefixing to **all** elements with existing IDs, not just those in the list
- Uses some sets instead of lists for checking id existence for o(logn) checks over o(n) checks
- Removes randomness from auto id generation and just start counting up from zero to determine auto id suffixes. This should help us make book builds more deterministic as well
- Added onto `test_from_formatter_to_adapter` (See: https://github.com/openstax/cnx-epub/pull/167/files#diff-38406bf6e5b6bc25a346d89b5c4d6199R703) to check that modules with duplicate IDs that get thrown into the `adapt_single_html` function are not duplicated afterwards.
- Adjusts regression tests to match and removes seeded randoms from the tests, since they are no longer needed (and made one or two tweaks to utility functions)
- Partially resolves a race condition that existed with `SingleHTMLFormatter->includes` in order to get tests to pass consistently

### One regression that I ignored, but I want to discuss:
Elements that were not already on the list of elements to prefix, but now receive prefixing which also already started with `auto_` are given a second `auto_` prefix. The regression test perhaps suggests this is something that should occur. See: https://github.com/openstax/cnx-epub/pull/167/files#diff-93849c0776f90f81b2d971f5cb915dfeR272 (`auto_apple_auto_apple`). Should we add a special case to not prefix ids that already begin with `auto_` to avoid this?

### DIFF
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -300,8 +300,8 @@ class SingleHTMLFormatter(object):
         self._build_binder(self.binder, self.body)
         # Fetch any includes from remote sources
         if not self.included and self.includes is not None:
-            with ThreadPoolExecutor(max_workers=self.threads) as e:
-                for match, proc in self.includes:
+            for match, proc in self.includes:
+                with ThreadPoolExecutor(max_workers=self.threads) as e:
                     for elem in self.xpath(match):
                         e.submit(proc, elem)
             self.included = True
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -9,7 +9,6 @@ from __future__ import unicode_literals
 import hashlib
 import json
 import logging
-import random
 import sys
 from io import BytesIO
 
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -109,11 +109,16 @@ class HTMLFormatter(object):
 
         def next_free_auto_id_generator():
             next_free_auto_id_generator.new_id_count = 0
+
             def inner():
-                auto_id = 'auto_{}_{}'.format(document_id, next_free_auto_id_generator.new_id_count)
+                auto_id = 'auto_{}_{}'.format(
+                    document_id,
+                    next_free_auto_id_generator.new_id_count)
                 next_free_auto_id_generator.new_id_count += 1
                 while auto_id in existing_ids:
-                    auto_id = 'auto_{}_{}'.format(document_id, next_free_auto_id_generator.new_id_count)
+                    auto_id = 'auto_{}_{}'.format(
+                        document_id,
+                        next_free_auto_id_generator.new_id_count)
                     next_free_auto_id_generator.new_id_count += 1
                 return auto_id
             return inner
@@ -138,8 +143,10 @@ class HTMLFormatter(object):
             'blockquote', 'q', 'code', 'pre', 'object', 'img', 'audio',
             'video',
         ]
-        elements_xpath = '|'.join(['.//*[local-name() = "{}"]'.format(elem, elem)
-                                   for elem in elements_need_ids])
+        elements_xpath = '|'.join([
+            './/*[local-name() = "{}"]'.format(elem, elem)
+            for elem in elements_need_ids
+        ])
 
         data_types_need_ids = [
             'equation', 'list', 'exercise', 'rule', 'example', 'note',
--- a/cnxepub/formatters.py
+++ b/cnxepub/formatters.py
@@ -143,7 +143,7 @@ class HTMLFormatter(object):
             'video',
         ]
         elements_xpath = '|'.join([
-            './/*[local-name() = "{}"]'.format(elem, elem)
+            './/*[local-name() = "{}"]'.format(elem)
             for elem in elements_need_ids
         ])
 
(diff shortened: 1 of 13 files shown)


tomjw64:
Don't know why codecov isn't putting its report in the pr status, but here is the link: https://codecov.io/gh/openstax/cnx-epub/pull/167

cnxepub/formatters.py
```diff
@@ -101,48 +101,70 @@ def _generate_ids(self, document, content):
...
+        elements_need_ids = [
             'p', 'dl', 'dt', 'dd', 'table', 'div', 'section', 'figure',
             'blockquote', 'q', 'code', 'pre', 'object', 'img', 'audio',
-            'video', 'aside', 'ol'
-            ]
-        elements_xpath = '|'.join(['.//{}|.//xhtml:{}'.format(elem, elem)
-                                  for elem in elements])
+            'video',
+        ]
+        elements_xpath = '|'.join(['.//*[local-name() = "{}"]'.format(elem, elem)
```
rnathuji: I think this can just be `.format(elem)` now

rnathuji:
Minor comment - I think there are a couple of additional places we can cleanup use of `random` with your changes:
* [formatters.py](https://github.com/openstax/cnx-epub/blob/73e6baf7bd6a66b7466bcddea1a767e16caf16d8/cnxepub/formatters.py#L12)
* [test_single_html.py](https://github.com/openstax/cnx-epub/blob/73e6baf7bd6a66b7466bcddea1a767e16caf16d8/cnxepub/tests/scripts/test_single_html.py#L77-L78)

rnathuji:
@tomjw64 - To your question of whether to add a special case to not prefix ids that already begin with `auto_`, is there a known scenario where we'd expect that to occur, or is this more of a hypothetical based on the existing tests? If the latter, I wonder if passing it through without prefixing may not cause issues later in `fix_generated_ids` where it does the `split('_', 2)[-1]` on ids that begin with `auto_` since it may get garbled in an undesirable way at that point otherwise (?). It at least seems clearer to me to treat everything the same w/o any special casing, but I also probably don't have a complete understand of all of the implications so that opinion may not be worth much.

tomjw64:
@rnathuji My issue is that I have no idea why an `auto_` prefixed id would already be there, and if there is a reason, is it some historical issue we need to be aware of? I'm leaning towards not special casing it, and letting it pass through as well.

tomjw64:
Correct, only items in the whitelist were getting prefixed with auto. Now all items in the whitelist will be prefixed. I believe all the prefixes are removed when we adapt the book back into a binder (i.e. before rex gets it), but I could be wrong. I would be interested to know if our new web hosting pipeline output contains a bunch of auto_ prefixes

rnathuji:
FWIW (I had the files from build `20200708.184941` handy in a local directory), there seem to only be a few instances of `auto_` in a handful of pages:

```
$ grep -Rl "auto_" . | grep xhtml
./55931856-c627-418b-a56f-1dd0007683a8@4.5:b6f86a79-3398-4a84-8726-2f896582e2c9.xhtml
./e42bd376-624b-4c0f-972f-e0c57998e765@8.6:8eb77899-82c9-4c81-8219-8c14291ae0d4.xhtml
./e42bd376-624b-4c0f-972f-e0c57998e765@8.6:6c029c67-94e7-4533-adaf-433e545be8ba.xhtml
./e42bd376-624b-4c0f-972f-e0c57998e765@9.1:8eb77899-82c9-4c81-8219-8c14291ae0d4.xhtml
./8d04a686-d5e8-4798-a27d-c608e4d0e187@29.3:d1df5bf8-e6f7-4e8c-a2af-8b3e8ff58e32.xhtml
./e42bd376-624b-4c0f-972f-e0c57998e765@9.1:6c029c67-94e7-4533-adaf-433e545be8ba.xhtml
```

TomWoodward:
> > Removes randomness from auto id generation and just start counting up from zero to determine auto id suffixes. This should help us make book builds more deterministic as well
> 
> I think that these additional ids might actually make REX's highlighting harder. Were only the whitelist elements getting the `auto_*` id attributes on them?
> 
> /cc @TomWoodward what do you think of the content having fewer (maybe none) id attributes that begin with `auto_`? I recall you filter them out for highlighting reasons because they are so unreliable, right?

rex assumes that ids are auto generated if they are purely numeric, we ignore these and the ids which start with `term` when saving highlights, we don't have any logic looking for the word `auto`

philschatz:
Ah, in that case (REX does not filter out the `auto_` prefixed ids) it seems dangerous to include more of them, especially if the id could easily be on a different element in the future. That would cause a students highlight to move on them unexpectedly.

What do you folks think of removing auto-generated ids altogether? Especially since they are not guaranteed to be stable.

tomjw64:
I would note it could also be dangerous to remove ids without working with the styles team first to make sure they are not relying on them for recipes.

tomjw64:
Built business law with this change. I see no instances of `auto_` in the disassembled content. However it is a small book...

PR #170, opened 2020-12-09
initial gh actions config for running unit tests

m1yag1:


rnathuji (commented):
Very cool! I did notice the new config doesn't incorporate the `pep8` linting from the older travis steps. Was that on purpose?

m1yag1:
oh good catch! that was not on purpose. I got ahead of myself. i was too excited.

m1yag1:
i'll add that now.

rnathuji (approved):
Much faster than Travis! 😃

m1yag1:
@rnathuji I added the linting step you noticed was missing. On another note, mock should be installed already in the python 2.7 run but it wasn't for some reason. That's why I've added now and doesn't seem to affect the 3.6 env either. I debated removing python 2.7 as we don't run it in production. What do you think?

rnathuji:
Isn't publishing still on python2? It has `cnx-epub` as a dependency.

m1yag1:
Oh you know what i'm getting my wires crossed 😅 . I've been working on cnx-press in cnx-deploy and was thinking of cnx-press. disregard my last.

PR #176, opened 2021-07-05
Prefix page IDs with `page_` when formatting

rnathuji:


rnathuji:
I'm going to go ahead and merge this with the +1 to keep the issue progressing, but if you get a chance to take a look and see any red flags, definitely comment here or in the [parent issue](https://github.com/openstax/cnx/issues/1514) @philschatz.

philschatz (commented):
Thanks! It took a little to figure out what was going on but I don’t think there’s a way to make it much nicer besides explicitly accounting for every way we prefix ids.

PR #177, opened 2021-07-19
Handle bad tag values during exercise annotation

rnathuji:


cnxepub/formatters.py
```diff
@@ -426,7 +429,77 @@ def exercise_callback_factory(match, url_template,
...
+            # No valid page UUIDs in exercise data
+            return
+
+        if parent_page_uuid in candidate_uuids:
+            target_module = parent_page_uuid
+        else:
+            # Use an UUID in the intersection of page UUIDs and tag UUIDs
+            # This is somewhat arbritrary, but if we hit this scenario with
+            # more than one UUID in the set things have gone pretty
+            # unexpectedly
```
philschatz, line 493: Just from reading this code it seems the formatter has the whole book inside self.root. If so, could you run one more check and throw if the `auto_{target_module}_{feature}` does not exist? That way it will be easier to debug than trying to find out why validation failed later on.

cnxepub/formatters.py
```diff
@@ -426,7 +429,77 @@ def exercise_callback_factory(match, url_template,
...
+        # we make sure the parent page UUID it's included in candidate_uuids
+        maybe_feature = parent_page_elem.find(
+            './/*[@id="{}"]'.format("auto_%s_%s" % (parent_page_uuid, feature))
+        )
+        if (maybe_feature is not None):
+            candidate_uuids.add(parent_page_uuid)
+
+        if len(candidate_uuids) == 0:
+            # No valid page UUIDs in exercise data
+            return
```
philschatz, line 485: I can't remember what @oscryan mentioned last week. I'm still worried about silently failing (not including the context information) when a `context-cnxfeature:` tag is set on an exercise but no valid module UUID could be found.

philschatz (changes requested):
Thanks for including the comments! They'll definitely help a lot the next time someone is in this codebase.

cnxepub/formatters.py
```diff
@@ -426,7 +429,77 @@ def exercise_callback_factory(match, url_template,
...
+
+        # There may be multiple `context-cnxmod:{uuid}` tags, each of which
+        # could be the parent page for this exercise, a different page in this
+        # book, or even invalid altogether. We'll prefer the first, fallback
+        # to the second, and not annotate in the last case.
+
+        # Determine parent page for this exercise
+        parent_page_uuid = None
+        parent_page_elem = None
+        for parent in elem.iterancestors():
```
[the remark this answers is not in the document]
rnathuji: Yeah, let me try using `find` here.

Just FYI - I was having having seg fault issues when I tried using `xpath`. I think these occurred due to thread safety issues, but [from here](https://lxml.de/FAQ.html#can-i-use-threads-to-concurrently-access-the-lxml-api) it wasn't obvious to me whether this would be expected or not 🤷‍♂️ . Curious if you've had experience with this type of problem with threads and `lxml` in the past?

cnxepub/formatters.py
rnathuji: Ah, so something like `find()` up to the body and search for the ID before inserting it? Should we just do this in general as a validation check before the code below that sets the context? (e.g. in case we select the current page UUID in spite of it not being there which I realize now the current code path would actually allow if it's in the tags)

Also, just as FYI: We can add a throw here, but one thing I realized is even if / where errors occur in these exercise processing threads they get silently tossed today and won't actually break `assemble` per [this recent bug](https://github.com/openstax/cnx/issues/1635) I created.

cnxepub/formatters.py
rnathuji: Ah, right. I wasn't clear on the best assumption to make here as far as whether it's never valid or a corner case that can occur. We can `raise` here if the former is true, though.

rnathuji: Per discussion with @oscryan , it is possible for these scenarios to occur (for example as content is edited and the feature ID no longer matches the content). However, instead of creating a hole for this edge case, we'll `raise` here and if it's really a valid scenario to keep the exercise w/o context, it will trigger us to remove the `cnxfeature` tag from the exercise.

cnxepub/formatters.py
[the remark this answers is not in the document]
rnathuji: @philschatz - Actually, I remember now why I went this route in the first place. I seemed to me like `find()` only supports searching down the tree (I'm not really clear on the subset of xpath supported by `ElementPath` TBH) and I hit the `xpath` seg fault issues with that approach. Can you point me to the appropriate incantation / docs if I'm wrong about the `find()` limitation 😅 ? Otherwise, perhaps I can take a deeper look at the `xpath` segfaults to see if I can ferret out the root cause.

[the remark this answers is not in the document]
rnathuji: Actually, never mind. [This change](https://github.com/openstax/cnx-epub/pull/177/commits/f8c2c5ce92f9a1dbbbd7d310bd04103fc7bd83d3) seems to work fine using `xpath`. I think the query I was doing when I saw seg faults was down tree, but I tested this with real books and it doesn't seem to keel over.

cnxepub/formatters.py
rnathuji: Addressed with [this change](https://github.com/openstax/cnx-epub/pull/177/commits/5cefef609abb33ced1d527d668985df398cb91ae).

cnxepub/formatters.py
rnathuji: I went ahead and [cleaned up](https://github.com/openstax/cnx-epub/pull/177/commits/b7974c51ae7916431789333b0f2b95882d85df26) the parent page check to discard that UUID if we find the feature doesn't exist there, and added a [general validation](https://github.com/openstax/cnx-epub/pull/177/commits/a5312c8271f85982f51a2e76587d7aa2e7554be2) of the final context determined. This should ensure that any invalid feature link will cause `assemble` to fail once [this bug](https://github.com/openstax/cnx/issues/1635) is fixed as well.

rnathuji:
I've addressed all of the review comments and unit tests are passing. However, testing with real book content and `neb` CLI I'm seeing seg faults during `assemble`. I suspect this has to do with use of `xpath` and threads in the callback function that I stumbled upon earlier, but I'll dig more.

cnxepub/formatters.py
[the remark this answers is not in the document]
philschatz: Oh my thanks for digging through to find the root of the multithreaded problem and fixing it (using both `.xpath()` and `.find()`

philschatz (approved):

PR #179, opened 2021-08-11
refactor exercise template into file & setup unit tests

kswanson33:


cnxepub/formatters.py
```diff
@@ -575,56 +576,27 @@ def _replace_exercises(elem, page_uuids):
...
-            elem = parent
-            parent = elem.getparent()
-
-        parent.remove(elem)  # Special case - assumes single wrapper elem
         for child in nodes:
-            parent.append(child)
+            parent.insert(parent.index(elem), child)
+        parent.remove(elem)  # Special case - assumes single wrapper elem
 
     xpath = '//xhtml:a[contains(@href, "{}")]'.format(match)
```
kswanson33, line 583: @phil We worked on changing this line (xpath selection & replacement) to fit with the new cnxml exercise format. But it turns out that, with the above changes to how new elements are injected, the same selector can still be used. What would be best: trying to come up with a way to change the selector (this was the bug we were still stuck on when we paired the first time), or keeping it as is?

cnxepub/templates/exercise_template.xhtml.jinja
```diff
@@ -0,0 +1,56 @@
...
+        {% for solution in question.collaborator_solutions %}
+        <div data-type="solution" data-solution-source="collaborator" data-solution-type="{{ solution.solution_type }}">
+            {{ solution.content_html }}
+        </div>
+        {% endfor %}
+        {% for solution in question.community_solutions %}
+        <div data-type="solution" data-solution-source="community" data-solution-type="{{ solution.solution_type }}">
+            {{ solution.content_html }}
+        </div>
+        {% endfor %}
```
kswanson33, line 53: @phil you mentioned putting the solutions inside a list, but I wonder if it may be uncommon for these questions to each have more than one solution; if that's the case, it might be better to leave them as-is and let the recipe handle formatting multiple solutions

cnxepub/tests/test_formatters.py
```diff
@@ -1376,3 +1376,263 @@ def test_annotate_autdetectparent(self, requests_get):
...
+                ],
+                "uuid": "a6581f0e-529b-4eaf-b174-563c7c49831b",
+                "group_uuid": "2c185140-d8dd-4116-83fa-f17783d0c2e3",
+                "number": 25499,
+                "version": 3,
+                "uid": "25499@3",
+                "published_at": "2021-08-04T18:43:46.007Z",
+                "nickname": "contmath98",
+                "url": "test-url", # Added in formatters
+                "required_context" : {  # Added in formatters
```
philschatz, line 1404: These comments are easy to miss. Would it help to reduce the size of this JSON by removing unused fields (like `"images": []` and `"authors":`?

Also, it might help to inject the `url` and `required_context` fields in as part of the test.

kswanson33: I agree it would be good to inject these as part of the test; I'm going to make sure `url` is covered the same way `required_context` is covered in earlier tests. The reason why I don't do that here is to be more modular; URL and required context come from that `exercise_callback_factory` function, which at this point is doing a lot of different things, and I wanted to make a separate function to interface with the template so that the template can be tested on its own. Let me know what you think.

cnxepub/templates/exercise_template.xhtml.jinja
```diff
@@ -0,0 +1,56 @@
...
+        data-type="question"
+        data-is-answer-order-important="{{ question.is_answer_order_important }}"
+        data-formats="{{ ' '.join(question.formats) }}"
+    >
+        {% if question.stimulus_html %}
+        <div data-type="question-stimulus">{{ question.stimulus_html }}</div>
+        {% endif %}
+        <div data-type="question-stem">{{ question.stem_html }}</div>
+        {% if question.answers %}
+        <ol data-type="multiple-choice-options" type="a">
```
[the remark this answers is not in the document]
kswanson33: I'll change it to `answers`. Is `option` still a good data type for the individual `li`s?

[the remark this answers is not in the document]
philschatz: I think changing them to `answer` would be sell surprising. I'm just worried that a future dev won't know about the `"answer" -> "option"` translation and get confused looking at the exercise JSON and the assembled HTML.

cnxepub/templates/exercise_template.xhtml.jinja
```diff
@@ -0,0 +1,50 @@
+<div
```
philschatz, line 1: :thinking: I noticed that some of the data-type attributes contained their parent (`exercise-stimulus`) and others did not (`question`, `option`) and that got me thinking.

I'm not sure if there was a reason behind one vs another but for greppability I like the hyphenated types. It also helps with validation to distinguish between the in-CNXML exercises and the injected ones because the markup is very different. Would something like this work?

```html
<div 
    data-type="injected-exercise"
    data-injected-from-nickname="{{ data.nickname }}"
    data-injected-from-version="{{ data.version }}"
    data-injected-from-url="{{ data.url }}"
    data-tags="{{ ' '.join(data.tags) }}"
    data-is-vocab="{{ data.is_vocab }}">
    {% if data.required_context -%}
    <div data-type="exercise-context"
        data-context-module="{{ data.required_context.module }}"
        data-context-feature="{{ data.required_context.feature }}">
        <a class="autogenerated-content"
        href="#{{ data.required_context.ref }}">[link]</a>
    </div>
    {% endif -%}
    {% if data.stimulus_html -%}
    <div data-type="exercise-stimulus">{{ data.stimulus_html }}</div>
    {% endif -%}
    {% for question in data.questions -%}
    <div 
        data-type="exercise-question"
        data-is-answer-order-important="{{ question.is_answer_order_important }}"
        data-formats="{{ ' '.join(question.formats) }}">
        {% if question.stimulus_html -%}
        <div data-type="question-stimulus">{{ question.stimulus_html }}</div>
        {% endif -%}
        <div data-type="question-stem">{{ question.stem_html }}</div>
        {% if question.answers -%}
        <ol data-type="question-answers" type="a">
            {% for answer in question.answers -%}
            <li data-type="question-answer" data-id="{{ answer.id }}" data-correctness="{{ answer.correctness }}">
                <div data-type="answer-content">{{ answer.content_html }}</div>
                {% if answer.feedback_html %}<div data-type="answer-feedback">{{ option.feedback_html }}</div>{% endif %}
            </li>
            {% endfor -%}
        </ol>
        {% endif -%}
        {% for solution in question.collaborator_solutions -%}
        <div data-type="question-solution" data-solution-source="collaborator" data-solution-type="{{ solution.solution_type }}">
            {{ solution.content_html }}
        </div>
        {% endfor -%}
        {% for solution in question.community_solutions -%}
        <div data-type="question-solution" data-solution-source="community" data-solution-type="{{ solution.solution_type }}">
            {{ solution.content_html }}
        </div>
        {% endfor -%}
    </div>
    {% endfor -%}
</div>
```

philschatz (approved):

cnxepub/formatters.py
philschatz: Oh, I forgot about this one. I _think_ keeping this here will work but I'm not 100% sure. If you're able to use this code to build multiple books we can verify this code still builds the other books that uses embedded exercises (I think it's microbiology)

philschatz (approved):

PR #183, opened 2021-11-11
Remove HTML-parsing fallback

philschatz:
The MathML namespace is currently being stripped in all content because `html` was already read once

Alternate solution: https://github.com/openstax/output-producer-service/pull/410

Example HTML where the [MathML namespace is stripped](https://openstax.org/apps/archive/20210823.155019/contents/031da8d3-b525-429c-80cf-6c8ed997733a@23.42:c82515ee-02ef-4e15-9b4f-52e67617359c.xhtml)


TylerZeroMaster (approved):
I like this fix better because, if I understand correctly, it is closer to the source of the problem.

PR #185, opened 2022-01-18
remove html test of non existing html fallback

therealmarv:
The html parsing fallback was removed in https://github.com/openstax/cnx-epub/pull/183

This removes the html test which makes no sense to test anymore.


therealmarv:
I will close this PR because it is merged in the ampersand one.

PR #187, opened 2022-10-17
:fire: Remove dead code

philschatz:
Running code coverage while building all the books resulted in these lines not being covered. See [enki coverage report](https://philschatz.com/enki-coverage/coverage/)

# TODO

- [x] Pin PyICU to 2.10.1


cnxepub/formatters.py
```diff
@@ -535,8 +460,7 @@ def _replace_exercises(elem, page_uuids):
 
 
 def render_exercise(exercise):
-    if len(exercise['items']) != 1:
-        raise Exception('Exercise "items" array is nonsingular')
+    assert len(exercise['items']) is 1, 'Exercise "items" array is nonsingular'
```
TylerZeroMaster, line 463: [Using `is` to compare to literals "works" on CPython by accident](https://bugs.python.org/issue34850)

TylerZeroMaster (changes requested):
Everything else looks good to me.

TylerZeroMaster (approved):

